PR #173, opened 2011-04-13
Fixing spelling mistake (runing -> running)

scompt:


sant0s:
@scompt @mbknor Play 1.5.0 does not include this fix, so wondering if it has ever made it to the master.

xael-fry:
@sant0s I open #1228 to fix it

PR #405, opened 2011-12-13
[#1310] Fixed package-info.java loading when running from precompiled cl...

jkot:
I'm not sure you want to fix it exactly this way but at least it will hopefully save you some time localizing the root cause. See the ticket where I attached a sample application that can be used to test the problem.

### DIFF
--- a/framework/src/play/classloading/ApplicationClasses.java
+++ b/framework/src/play/classloading/ApplicationClasses.java
@@ -260,8 +260,12 @@ public class ApplicationClasses {
         }
 
         public boolean isClass() {
-            return !name.endsWith("package-info");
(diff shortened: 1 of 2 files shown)


tazmaniax:
Hi @jkot , I'm getting the same problem again with Play v1.3.1-56-gac02587 on OracleJDK8 when I run using "play-13x run -Dprecompiled=true". What was the reasoning behind your changes in this pull request and do you have any tips on how to find the root cause. I've stepped through ApplicationClassloader.loadApplicationClass(String name) in both non-precompiled and precompiled modes but I can't see any effective difference in the loaded class for models/package-info.java. thx

PR #635, opened 2013-08-14
Fixed field name detection in 'attachment' Content-Disposition headers

svintus:
This fixes an issue I was seeing when uploading a file with `Content-Disposition: attachment;` where the uploaded file wouldn't bind to FileUpload. I found that the problem went away if I set `Content-Disposition` to `form-data`. From there I was able to track the issue to the line in question and fix it.


xael-fry:
[Lighthouse #2009](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2009-field-name-detection-in-attachment)

xael-fry:
Thanks

PR #751, opened 2014-04-26
Security bug : leak private data of last user

ilinz:
Current thread does not clear, so the exception thrown by the parseRequest method can leak private data of last user.
Request headers with non-numeric port will cause error by the parseInt method and lead to problems.

### DIFF
--- a/framework/src/play/server/PlayHandler.java
+++ b/framework/src/play/server/PlayHandler.java
@@ -98,10 +98,12 @@ public class PlayHandler extends SimpleChannelUpstreamHandler {
 
             // Plain old HttpRequest
             try {
-                final Request request = parseRequest(ctx, nettyRequest, messageEvent);
-
+                //Reset request object and response object for the current thread.
+                Http.Request.current.set(new Http.Request());
                 final Response response = new Response();
                 Http.Response.current.set(response);
+
+                final Request request = parseRequest(ctx, nettyRequest, messageEvent);
 
                 // Buffered in memory output
                 response.out = new ByteArrayOutputStream();


alexanderkjall:
Would it be possible to apply for a CVE for this issue? That would greatly simplify for us to track the issue in our organization. And it would also be great if you could publish a list of affected versions somewhere, (or maybe you have and I just can't find it).

asolntsev:
@alexanderkjall Sorry, I don't understand your question. What exactly should we do? What means "apply for CVE"?

I guess affected version is 1.2.x and all previous versions.

alexanderkjall:
Hi, sorry that I maybe came of as a bit aburpt, I didn't manage to find your page with security information (this one https://www.playframework.com/security/vulnerability/20151230-SessionHijack ) and was stressed.

Regarding your question, a CVE is a tracking number for security problems. It's used both as a way to easily identify and talk about security problems, and also there is automated tooling that can notify you about security issues in your dependencies (for example: https://www.owasp.org/index.php/OWASP_Dependency_Check ).

You as a project can apply for CVE's here: https://docs.google.com/forms/d/e/1FAIpQLSeiY7ldJAx-fjU6eSnXDaX5TB--L1ujCQpmGAKnqBSJOcBShw/viewform (link came from here: https://cve.mitre.org/cve/request_id.html ).

Best regards.

asolntsev:
@alexanderkjall Thank you for the clarification. But this problem existed very long time ago - in Play 1.2.x. This is very old version! Why do you still need it?

alexanderkjall:
@asolntsev because i work in an organisation that built itself into a corner and have managed to get out of it T_T.

Applying for a CVE isn't important for me now, but it would be helpful in the future if security problems got logged, since that would enable our automated tooling to detect and alert us.

PR #775, opened 2014-07-08
[#1836] fix continuation issues after asm/javaflow update

gbevin:
I've been porting our LiveRebel application to Play 1.3 with the intent of supporting Java 8 execution and development. These are all the changes that I had to make to get it to compile and start up.


gslowikowski:
Hi. I'm building patched ``commons-javaflow`` from sources to generate ``-source`` and ``-javadoc`` artifacts. Almost all unit tests fail with ``java.lang.IllegalStateException: Undefined label used`` exception. Should I worry or tests just need fixing after moving to asm 5?

PR #778, opened 2014-07-13
[#1849] Exceptions from controllers are not detected in FunctionalTests

hsch:
- Fix for #1849
- Includes test cases that pass only with the fix, and fail without
- Made one test more robust with respect to timing
- Tests "continuations", "Promises", and "CRUD" are still unreliable, but hopefully not related to this change


xael-fry:
Rebase this PR in #937

PR #800, opened 2014-09-16
[#1873] Feature/groovy templates performance

cbxp:
"play precompile" is very slow on big projects. The problem is in method `GroovyTemplateCompiler.source` that performs genormous amount of `String.replaceAll` operations. Removing every `String.replaceAll` helped to boost performance. Our applications work fine with play-1.3.x.

http://play.lighthouseapp.com/projects/57987-play-framework/tickets/1873-boost-groovytemplatecompiler-performance

### DIFF
--- a/framework/src/play/templates/GroovyTemplateCompiler.java
+++ b/framework/src/play/templates/GroovyTemplateCompiler.java
@@ -1,18 +1,15 @@
 package play.templates;
 
 import groovy.lang.Closure;
-import java.io.PrintWriter;
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.regex.Pattern;
 import play.Play;
 import play.exceptions.TemplateCompilationException;
 import play.templates.GroovyInlineTags.CALL;
+
+import java.io.PrintWriter;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 /**
  * The template compiler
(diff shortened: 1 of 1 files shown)


flybyray:
Not happy with this default, because you distribute a module named 'console' with play which relays on it enabled:
https://github.com/playframework/play1/blob/1.3.x/modules/console/app/views/console/index.html#L50
https://github.com/playframework/play1/blob/1.3.x/modules/console/app/views/console/index.html#L53

Now i have to distribute/introduce a configuration change 'groovy.template.check.scala.comptatibility=true' which makes no sense at all, because it is a native java webapp not using any modules with scala runtimes or something similar.

@cbxp: What should I expect from your statement "Our applications work fine"?

xael-fry:
The module console is in the repo since more than 5 years, but if you don't include it it must not affect your project and even if you include it I didn't where this module need the 'scala' parameter? And I dont understand why you have to introduce this param if you don't  use scala?

If you have any issue please create an issue and we will have a look on it.

flybyray:
Here you are.
https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1956-adding-applicationclass-with-wrong-names

xael-fry:
Ok thanks, I will have a look on it.

flybyray:
Thanks a lot.
Maybe you will have a look at https://play.lighthouseapp.com/projects/57987/tickets/1669-seemingly-random-classnotfoundexception-on-first-request too.
You will see the problem is quite similar. Groovy will add MetaClass information to some special classes. For example use the play crud module, make the ObjectType abstract and use a derived ObjectTypeInner class to instantiate the objects (only once). This pattern can be usefull to create cloned objects (that maybe quicker for some applications). But if you do so, you will see that groovy wants to load a CRUD$ObjectType$ObjectTypeInnerBeanInfo and CRUD$ObjectType$ObjectTypeInnerCustomizer class.

asolntsev:
Hi!
I have found cause of this problem. Here is a description: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1669-seemingly-random-classnotfoundexception-on-first-request#ticket-1669-3

flybyray:
I will check it again with #886 . Regardless of #886 I have to enable 'groovy.template.check.scala.comptatibility' to make groovy template code work as expected and described in play documentation and showed in https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1956-adding-applicationclass-with-wrong-names

PR #805, opened 2014-10-04
Make module name extraction consistent

angryziber:
Always assume that version number is after the last dash '-', not sometimes after the first one.

This will resolve module names like play-selenide-1.2.3 correctly in both cases.

Current code works correctly in case of MODULES env variable, but wrongly in case of listing the modules dir.


asolntsev:
Closing this pull request.

PR #819, opened 2014-11-19
Normalize number separators while binding

leon:
If using a locale which has comma instead of dot as a separator the bindings would cause everything after the decimal to get dropped.

By using the current locale to normalize the group and decimal chars, we allow the bindings to work with all languages


xael-fry:
Could you created and issue on [lighthouse](http://play.lighthouseapp.com/projects/57987-play-framework/overview)?

PR #820, opened 2014-11-23
[#1979] Performance optimization

asolntsev:
Performance optimization of some play methods.
The most important is the first commit by Nov 11, "avoid pointless loading the whole class definition" which speeds up our applications by ~15%.

### DIFF
--- a/framework/src/play/classloading/ApplicationClassloader.java
+++ b/framework/src/play/classloading/ApplicationClassloader.java
@@ -213,9 +213,7 @@ public class ApplicationClassloader extends ClassLoader {
             return null;
         }
         try {
-            ByteArrayOutputStream os = new ByteArrayOutputStream();
-            IOUtils.copyLarge(is, os);
-            return os.toByteArray();
+            return IOUtils.toByteArray(is);
         } catch (Exception e) {
             throw new UnexpectedException(e);
         } finally {
--- a/framework/src/play/classloading/ApplicationClassloader.java
+++ b/framework/src/play/classloading/ApplicationClassloader.java
@@ -467,6 +467,10 @@ public class ApplicationClassloader extends ClassLoader {
      */
     public Class getClassIgnoreCase(String name) {
         getAllClasses();
+        Class<?> aClass = loadApplicationClass(name);
+        if (aClass != null) {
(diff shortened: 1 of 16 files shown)


asolntsev:
the recent commit "speed up method getClassIgnoreCase(): get class from Map instead of i… " also improves performance gradually. Not it's spend quite a lot of CPU to iterate all classes multiple times. Getting them from Map is much faster.

asolntsev:
Here is ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1979-improve-performance-of-play-13x

framework/src/play/classloading/ApplicationClassloader.java
```diff
@@ -470,11 +468,13 @@ public int compare(Class o1, Class o2) {
...
-            if (c.name.equalsIgnoreCase(name) || c.name.replace("$", ".").equalsIgnoreCase(name)) {
-                if (Play.usePrecompiled) {
-                    return c.javaClass;
-                }
-                return loadApplicationClass(c.name);
+            if (c.name.equalsIgnoreCase(name)) {
+                return Play.usePrecompiled ? c.javaClass : loadApplicationClass(c.name);
+            }
+        }
+        for (ApplicationClass c : Play.classes.all()) {
```
xael-fry: Any response for this question?

asolntsev: @tazmaniax @xael-fry Sorry for ignoring your question.

The goal is to improve performance. Operations `equalsIgnoreCase` and especially `.replace("$", ".")` are relatively slow - they are executed many times in the loop. But in most cases it's not needed, because developers do write correct class names in HTML (like `@{MyController.myAction()}`).

So, now the algorithm is:
1. Try the **fastest method** that covers most of cases - `loadApplicationClass(name);`   (take class from Map, avoid looping)
2. Try the second **slower method** in loop - `c.name.equalsIgnoreCase(name)`
3. Try the **slowest method** `c.name.replace("$", ".")`.

xael-fry:
Thanks

asolntsev:
@xael-fry Cool! Thank you!

PR #822, opened 2014-11-29
Netty v4.0.24 support

tazmaniax:


tazmaniax:
hey @Fraserhardy , unfortunately I didn't get any further and I don't have any time right now to pick it up. We should have a catch up some time :)

flybyray:
There is a migration guide for a sample application https://netty.io/wiki/new-and-noteworthy-in-4.0.html#wiki-h3-41 it is a good starting point.
The document points out some fundamental changes. Some collide very heavy with the Play! usage of the netty library.
I provide a starting point. But with ssl some blocking occurs (may be some channel close event missing).
https://github.com/flybyray/play1/commit/e0726f4f60b89ad08ac3b57277e1f686514e96bf

PR #823, opened 2014-12-08
[#1894] Updated Javassist to v3.19.0 RC1 that resolves an exception when enhancing a class with a Java 1.8 lambda expression

tazmaniax:


xael-fry:
Merged ans pushed in master and 1.3.x.
Thanks

PR #827, opened 2015-01-06
[#1899] fix(PlayHandler): check if the channel because writing on it

xael-fry:


xael-fry:
Merged in master and 1.3.x.
Thanks

PR #830, opened 2015-01-15
[#1902] Use JAVA_HOME in getJavaVersion

gpgekko:
Moving `java_path()` out of application into utils so `getJavaVersion()` can use it too


xael-fry:
Thanks

PR #835, opened 2015-02-12
[#1916] Get Java version from application.conf

opyate:
```getJavaVersion()``` (introduced in 8f5e3e591baf50cbfda11722cd4195957c598d14) spawns a subprocess, which bumps the PID count, which breaks [Upstart](http://upstart.ubuntu.com/), the latter which expects Play to fork exactly none, once or twice, depending on which ```expect``` stanza you use.

### DIFF
--- a/framework/pym/play/application.py
+++ b/framework/pym/play/application.py
@@ -274,7 +274,7 @@ class PlayApplication(object):
         if application_mode == 'prod':
             java_args.append('-server')
 
-        javaVersion = getJavaVersion()
+        javaVersion = self.readConf('java.source')
         print "~ using java version \"%s\"" % javaVersion
         if javaVersion.startswith("1.7"):
             # JDK 7 compat
--- a/framework/pym/play/application.py
+++ b/framework/pym/play/application.py
@@ -218,7 +218,6 @@ class PlayApplication(object):
         return log_path
 
     def check_jpda(self):
-        self.jpda_port = self.readConf('jpda.port')
         try:
             s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             s.bind(('', int(self.jpda_port)))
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by xael-fry** [comment]
Could you create an issue?
Thanks

**Comment by gpgekko** [comment]
I understand the goal here, but this change would be very inconvenient. Perhaps we can add a command line argument to specify the Java version instead, and have it fall back to the current behavior if the argument isn't provided. That would be much more convenient for the majority of use cases.

**Comment by xael-fry** [comment]
@gpgekko  I agree, problem with this PR  is that "java.source"  version is not necessary the version of the running JVM

**Comment by opyate** [comment]
> @gpgekko I agree, problem with this PR is that "java.source" version is not necessary the version of the running JVM

I'll amend the PR.

What would you like me to call the flag? (e.g. ```jvm.version```)

**Comment by gpgekko** [comment]
Flag-only. As for the name, I'm not sure, there seem to be mixed conventions (`--pid_file` vs `--http.port`). Either way, the documentation and help text / command description will need to be updated to reflect the changes.

**Comment by gpgekko** [comment]
Perhaps change the title of the pull request for future reference. The rest looks fine to me.

**Comment by opyate** [comment]
:+1:

**Comment by gpgekko** [comment]
I don't think this got merged yet?

**Comment by opyate** [comment]
Those buttons are too close together.

File: framework/pym/play/application.py [hunk 1]
```diff
@@ -218,7 +218,6 @@ def log_path(self):
         return log_path
 
     def check_jpda(self):
-        self.jpda_port = self.readConf('jpda.port')
```
xael-fry: Why did you remove this line?

gpgekko: I think he might be referencing that in https://github.com/playframework/play1/pull/835#commitcomment-9746326

xael-fry: Ok saw it thanks

xael-fry:
Merged and pushed in master and 1.3.x.
Thanks

PR #838, opened 2015-02-15
[#1148] fix(tag): yesno should  evaluates NullObject as false

xael-fry:
[lighthouse #1148](http://play.lighthouseapp.com/projects/57987/tickets/1148-yesno-evaluates-nullobject-as-true)


xael-fry:
Thanks

PR #841, opened 2015-02-15
[#1919] fixes  Building modules with a space in path under windows fails

flybyray:


xael-fry:
Close in favor of #1020

PR #846, opened 2015-02-16
[#1923] Respect debugSQL of other configs

gpgekko:
If any of the databases want the logging to be on, don't let the others turn it off again.


xael-fry:
Merged in 1.3.x
Thanks

PR #828, opened 2015-01-13
[#1861] feat(validation): add new methods insert, remove, hasError

xael-fry:
[lighthouse #1861](http://play.lighthouseapp.com/projects/57987/tickets/1861-validation-error-manipulation)

PR #847, opened 2015-02-17
[#828] fixes: Module names can’t contain dash characters

flybyray:


aleksandy:
@flybyray, why can't a module name have a dash? What are the reasons for this?

PR #849, opened 2015-02-17
[#1908] fixes: @OnApplicationStart fails if you disable the JPAPlugin

flybyray:
I would prefer this solutions, because from a functional perspective it is essential that the play.Invoker.Invocation.execute() is called and not that a filter is found!


xael-fry:
Merged in 1.3.x.
Thanks

PR #850, opened 2015-03-01
[#1924] Sane c3p0 logging

swaldman:
play1 simultaneously uses c3p0 as its default Connection pool but sends all c3p0 logging to `/dev/null` (metaphorically speaking). That renders applications which experience database problems very difficult to debug.

c3p0's logging is unreasonably verbose at DEBUG/FINE/TRACE levels, but is reasonably quiet at INFO and very quiet at WARN or above, unless something really breaks. The library is intended to be logged at INFO. But it does emit an full configuration dump on pool initialization at INFO.

Presumably someone got annoyed by c3p0's logging and (insanely) just swept it away all at once.

This pull request restores c3p0 logging at WARN level. From the perspective of a guy who often gets c3p0 support requests, I'd prefer logging at INFO, so users can view and report their pool configurations. But given the extreme prejudice with which c3p0 logging was eliminated, I figure I should be grateful to even get logging at WARN. (I'd be glad to update this with a commit that logs at INFO instead, if you are amenable!)

### DIFF
--- a/framework/src/log4j.properties
+++ b/framework/src/log4j.properties
@@ -7,6 +7,7 @@ log4j.logger.DataNucleus=WARN
 log4j.logger.org.apache.commons.httpclient=ERROR
 log4j.logger.net.sf.oval.internal=ERROR
 log4j.logger.org.springframework=WARN
+log4j.logger.com.mchange=WARN
 
 # Console
 log4j.appender.Console=org.apache.log4j.ConsoleAppender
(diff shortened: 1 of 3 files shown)


xael-fry:
Could you create an issue in [lighthouse](http://play.lighthouseapp.com)?
Thanks

swaldman:
Sure. [Done](http://play.lighthouseapp.com/projects/57987-play-framework/tickets/1924).

angryziber:
FYI, we also set c3p0 logging to INFO on production, as some errors are extremely hard to find without it.

PR #852, opened 2015-03-03
[#1908] fix(job): @OnApplicationStart fails if no plugin with filter found (ex JPAPlugin  disabled)

xael-fry:
[lighthouse #1908](http://play.lighthouseapp.com/projects/57987-play-framework/tickets/1908)


xael-fry:
Merged in 1.3.x.
Thanks

PR #853, opened 2015-03-15
Make play.libs.WS.getString() and play.libs.WS.getQueryString() returnin...

Zork33:
...g same result on every call, by saving proper result in the private variable.

Story:

While switching from Play 1.2.7 to 1.3.0 found that secureSocial module fails to login thru Google OAuth2.  

Code study shown that play.libs.OAuth2 codes calls multiple times play.libs.WS. getQueryString() and play.libs.WS.getString(), but getString() is able to return response body string only once.  

In Play 1.2, for some unfound reаsons WS.getString() was returning a result only on a second call, and that is why whole thing was working.  But in Play 1.3 this behavior was somehow fixed, and getString() started to return the body string on a first call, as it suppose to be. And this made play.libs.OAuth2.Response not working, since this code was build around this bug.

As a fix, I suggest make getString() function to return the response body-string in any call, but saving the string in the private variable.

### DIFF
--- a/framework/src/play/libs/WS.java
+++ b/framework/src/play/libs/WS.java
@@ -527,7 +527,8 @@ public class WS extends PlayPlugin {
     public static abstract class HttpResponse {
 
         private String _encoding = null;
-
+        private String _getString = null;
+        private Map<String, String> _getQueryString = null;
         /**
          * the HTTP status code
          * @return the status code of the http response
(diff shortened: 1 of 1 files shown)


Zork33:
I see that test is failing, but cannot reproduce this error locally.  Is there any article how to setup proper local test environment for the Play code?

angryziber:
I suggest rebasing your changes to 1.3.x branch. Master seems to be quite different from it.

xael-fry:
see #959

PR #855, opened 2015-03-20
[#1934] Allow non-static controllers methods

angryziber:
Allow writing of non-static controller methods in Play 1.x, like in Play 2.4.x.

Non-static methods have lots of benefits, including:
- easier to extend/override
- easier to mock
- easier to unit-test in general

Change is fully backwards-compatible, as I have written unit tests for ActionInvoker before changing the method to make sure Scala support works the same way as before.

### DIFF
--- a/documentation/manual/controllers.textile
+++ b/documentation/manual/controllers.textile
@@ -30,25 +30,37 @@ import play.mvc.Controller;
  
 public class Clients extends Controller {
  
-    public static void show(Long id) {
+    public void show(Long id) {
         Client client = Client.findById(id);
         render(client);
     }
  
-    public static void delete(Long id) {
+    public void delete(Long id) {
         Client client = Client.findById(id);
         client.delete();
     }
  
 }
 
-Each public, static method in a Controller is called an action. The signature for an action method is always:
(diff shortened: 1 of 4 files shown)


arnzel:
+1 

That would be a great feature !

xael-fry:
Nice feature. Can you create an issue and add some example in **just-test-cases** sample project?
Thanks

xael-fry:
Ticket created [lighthouse #1934](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1934-allow-non-static-controllers-methods).
Can you add an Example and update **documentation/manual/controllers.textile** ?

Odilio:
Great!!

2015-04-15 9:34 GMT-03:00 Alexandre Chatiron <notifications@github.com>:

angryziber:
Thanks for creating an issue!
As for examples - just remove static keywords and it will work.

Before updating the documentation, I think it would be nice to create a new instance of controller only once, and run all befores/afters on the same instance. Current implementation creates a new instance on invocation of every method, so it is sort of experimental.

angryziber:
I have updated the docs and and changed code so that controller instance is created only once per request.

Unit tests cover that both old and new style works.

Is it good for a merge? :-)

angryziber:
Guys, it would be nice to get some feedback.

Probably, I have overdone with docs and we may stick to static by default, just mentioning that non-static is possible?

arnzel:
+1 for this feature

xael-fry:
Thanks

PR #856, opened 2015-03-31
[#1911] fix(FileService): check if channel is opened before using it

xael-fry:
[lighthouse #1911](https://play.lighthouseapp.com/projects/57987/tickets/1911-play-throw-exception-in-ssl-mode)


xael-fry:
Merged ans pushed in master and 1.3.x
Thanks

PR #862, opened 2015-04-26
[#1901] feat(lib): update to spymemcached 2.11.7

angryziber:
Previous version had some rare synchronization bugs that we caught in our application.
Please upgrade.


xael-fry:
Thanks

PR #865, opened 2015-04-27
[#1937] fix(router): make some improvement in router

xael-fry:


xael-fry:
Tests are ok on my computer on linux and Windows.
Merged it.

PR #868, opened 2015-04-27
[#1901] chore(lib): update some libs

xael-fry:


xael-fry:

PR #875, opened 2015-05-28
De-dupes list of plugins before load to avoid #176 and other issues.

johnw:
If a plugin with the same index and class name is specified multiple times, it is loaded only once.   This seems like a reasonable feature -- I can't think of a use case where loading the same plugin at the same priority multiple times would be desirable.

It also works around cases where Play.classloader.getResources("play.plugins") returns multiple copies of the plugin file.  https://play.lighthouseapp.com/projects/57987/tickets/176-app-playplugins-loaded-twice-conf-on-2-classpaths documents one case where this happened.    I've run into a similar issue occasionally with 1.3.1 on Win32 systems:   Play.classloader.getResources("play.plugins") will occasionally return ["/c:/app/path/conf/play.plugins", "/C:/app/path/conf/play.plugins"] and load everything twice.  I haven't been able to pin down why this happens.  It might be an issue with our environment or user error.   Regardless, de-duping the plugin list would band-aid problems like these.

### DIFF
--- a/framework/src/play/plugins/PluginCollection.java
+++ b/framework/src/play/plugins/PluginCollection.java
@@ -24,17 +24,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.net.URL;
-import java.util.AbstractCollection;
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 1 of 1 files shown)


xael-fry:
Merged and pushed in master, 1.3x and 1.4.x
Thanks

PR #878, opened 2015-05-29
[#1952] chore(librairies): Update async-http-client to v1.8.16

xael-fry:


xael-fry:
Merged and pushed in master and 1.3.x.
Thanks.

PR #879, opened 2015-06-02
[#1945] update html unit v2.16

xael-fry:
* need minimum of Java7 to use new lib htmlunit


xael-fry:
This PR requires at least ti use Java7 so we will need to use it in a new version 1.4.x where support for java 6 will be give up.

xael-fry:
Merged and pushed in master an 1.4.x

PR #880, opened 2015-06-02
[#1917] fix(ant): add "play libs" path to classpath to be able to call play.ant.PlayConfigurationLoadTask

xael-fry:
…l play.ant.PlayConfigurationLoadTask


xael-fry:
Thanks

PR #884, opened 2015-06-12
[#1278] fix(errorTemplate): need to check the type of exception befor…

xael-fry:
…e displaying the exception if (only available on PlayException)

[Lighthouse #1278](https://play.lighthouseapp.com/projects/57987/tickets/1278-500html-template-redering-issue)


xael-fry:
Thanks

PR #873, opened 2015-05-19
[#1946] chore(libraries): update from netty 3.9.4 to 3.10.3

xael-fry:
** remove some deprecated method call and used the new ones

[Play lighthouse #1758](https://play.lighthouseapp.com/projects/57987/tickets/1758)

PR #889, opened 2015-06-20
[#1959] chore(librairies): Update async-http-client to v1.9.27

flybyray:


flybyray:
This failed because tests run with old java.
[logfile](https://s3.amazonaws.com/archive.travis-ci.org/jobs/67661251/log.txt)
```
$ java -Xmx32m -version
java version "1.6.0_34"
```
The async-http-client library states in the [Migration Guide: Target JDK7, drop support for JDK5 and JDK6](https://github.com/AsyncHttpClient/async-http-client/blob/master/MIGRATION.md)

xael-fry:
Yes you need to upgrade travis, as i did in pull request #873, note that I still have some issues with this PR.

framework/src/play/test/FunctionalTest.java
```diff
@@ -187,27 +187,38 @@ public static Response POST(Request request, Object url, Map<String, String> par
...
-            } catch (FileNotFoundException e) {
-                throw new RuntimeException(e);
-            }
+            filePart = new FilePart(key, files.get(key));
             parts.add(filePart);
         }
 
-        MultipartRequestEntity requestEntity = new MultipartRequestEntity(parts.toArray(new Part[parts.size()]), new FluentCaseInsensitiveStringsMap()); 
-        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        MultipartBody requestEntity = null;
```
xael-fry: I am curious to know why you use **MultipartBody** and remove **MultipartRequestEntity**?

flybyray:
The problem is that MultipartRequestEntity was moved into another package and it is not compatible anymore. the BinaryTest and DataBinding Test will fail with it. I tested several solutions and inspected the async-http-code to see that developers introduced MultipartUtils with the static factory newMultipartBody in the same package.
1.8.*: https://github.com/AsyncHttpClient/async-http-client/blob/1.8.x/src/main/java/com/ning/http/multipart/MultipartRequestEntity.java
1.9.*: https://github.com/AsyncHttpClient/async-http-client/blob/1.9.x/src/main/java/com/ning/http/client/providers/jdk/MultipartRequestEntity.java

xael-fry:
Ok thanks, I think I will create a new branch 1.4.x, as we will drop support for java6

xael-fry:
Merged and pushed in 1.4.x

PR #893, opened 2015-06-25
[#1961] Fixes OpenID discovery issue.

X455u:
[Lighthouse #1961](https://play.lighthouseapp.com/projects/57987/tickets/1961-openid-discovery-fails-in-13x-in-some-cases)

### DIFF
--- a/framework/src/play/libs/OpenID.java
+++ b/framework/src/play/libs/OpenID.java
@@ -1,5 +1,6 @@
 package play.libs;
 
+import java.io.StringReader;
 import java.net.URI;
 import java.net.URLEncoder;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


framework/src/play/libs/OpenID.java
```diff
@@ -326,6 +330,16 @@ public static String discoverServer(String openid) {
         return server;
     }
 
+    private static Document getXml(String response) {
+        try {
+            InputSource source = new InputSource(new StringReader(response));
+            DocumentBuilder builder = XML.newDocumentBuilder();
```
xael-fry: I think we should specify the encoding ```source.setEncoding(encoding);```

X455u: According to this, specifying the encoding for the InputSource has no effect for character streams. http://docs.oracle.com/javase/6/docs/api/org/xml/sax/InputSource.html#setEncoding(java.lang.String)

xael-fry:
Thanks

PR #894, opened 2015-06-30
[#1947] [#1948] Fix PlayPlugin filter stacking.

johnw:
In 1.3.1, Plugin filters do not behave the way documentation suggests they should.  (For example, see javadoc for PlayPlugin::Filter.)    

Everything works normally as long as only one filter is defined (typically the one for JPAPlugin), but as soon as a second enters the picture everything goes off the rails. In Job, only the filter belonging to the highest priority plugin is executed.  In Invoker, actions are executed once per filter.  This does not seem like the intended behavior.

This patch makes plugins stack in the way documentation suggests they should.

### DIFF
--- a/framework/src/play/plugins/PluginCollection.java
+++ b/framework/src/play/plugins/PluginCollection.java
@@ -8,6 +8,7 @@ import play.classloading.ApplicationClassloader;
 import play.data.binding.RootParamNode;
 import play.db.Model;
 import play.exceptions.UnexpectedException;
+import play.libs.F;
 import play.mvc.Http;
 import play.mvc.Router;
 import play.mvc.results.Result;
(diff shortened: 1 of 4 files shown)


xael-fry:
Merged and pushed in 1.3.x and 1.4.x
Thanks

framework/src/play/plugins/PluginCollection.java
```diff
@@ -380,6 +411,35 @@ public void updatePlayPluginsList() {
...
+    public<T> F.Option<PlayPlugin.Filter<T>> composeFilters()
+    {
+      //Copy list of plugins here in case the list changes in the midst of doing composition...
+      //(Is it really necessary to do this?)
+      final List<PlayPlugin> pluginsWithFilters = new ArrayList<PlayPlugin>(this.getEnabledPluginsWithFilters());
+
+      if (pluginsWithFilters.isEmpty()) {
+        return F.Option.None();
+      } else {
+        final Iterator<PlayPlugin> itr = getEnabledPluginsWithFilters().iterator();
```
tazmaniax: Should this be using the cached list "pluginsWIthFilters" rather than calling "getEnabledPluginsWithFilters()" as per the comment above?

xael-fry: Yes you are right, fix it, thanks

PR #896, opened 2015-07-09
fix: allow zero-length blobs

sbeigel:
the fix for [#1564] breaks Blob binding, this fix resolves this by checking for an empty file name instead of a 0-size (which has resulted in a NPE)


sbeigel:
I created a lighthouse issue for this:

https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1969-allow-zero-length-blobs

xael-fry:
Thanks

PR #900, opened 2015-07-14
Remove method AutoFileItem.finalize()

asolntsev:
Avoid declaring finalize() methods when possible.

It makes GC work slower. GC puts all objects with "finalize" method to a separate "finalization queue", and GC collects them only during FullGC pause.

In Play! 1.3.x, `AutoFileItem` is used only in one place. It's easy to use `try/finally` and delete `AutoFileItem` files in finally block.

### DIFF
--- a/framework/src/play/data/parsing/ApacheMultipartParser.java
+++ b/framework/src/play/data/parsing/ApacheMultipartParser.java
@@ -439,17 +439,6 @@ public class ApacheMultipartParser extends DataParser {
         // ------------------------------------------------------ Protected methods
 
         /**
-         * Removes the file contents from the temporary storage.
-         */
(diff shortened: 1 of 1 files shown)


xael-fry:
Could you create an issue?

asolntsev:
@xael-fry There is issue: https://play.lighthouseapp.com/projects/57987/tickets/1966-avoid-declaring-finalize-methods

xael-fry:
Merged and pushed in
* master : f963059b260b04868c92fd0248e2f55f9a243b47
* 1.3.x : 07637155b76047937a82937bdac4e35ff9420df9
* 1.4.x. : 26a495422791e9799839ca336aa6d2a7a80a562c

Thanks

PR #902, opened 2015-07-30
[#1967] fix broken monitor statistics in status - make the code more future-proof as well

angryziber:
MonitorFactory.getRootMonitor().getBasicData() now returns one additional column (Instance), breaking index-based column access.

This patch now uses proper Monitor.getXX methods instead of error-prone indices and thus restores monitor status functionality in play 1.3.2


angryziber:
Guys, any attention to this?

xael-fry:
Sorry I was in holidays in july . I will work on it next week.


-------- Original Message --------

xael-fry:
Merged in 1.3.x and master.
Thanks

PR #903, opened 2015-08-09
[#1968] Update deployment.textile : Remove section about playapps.net

arnzel:
Remove section about playpps.net. It does no longer exist.


xael-fry:
Thanks

PR #906, opened 2015-08-26
* [#1972] fix(Router): Fix reverse routing of spaces to path not working because it doesn't replace "+" with "%20"

xael-fry:
[lighthouse 1972](https://play.lighthouseapp.com/projects/57987/tickets/1972)


xael-fry:
Merged and pushed in 1.3.x and master.
Thanks

PR #910, opened 2015-09-04
Fix jobs plugin

angryziber:
Currently, application status doesn't display Jobs manually shcheduled using Job.every().

Another issue is 'Scheduled jobs' and 'Waiting jobs' using different ways to display jobs: first one displays the class name directly, while the second uses toString().
Both should use the same method (toString), so that Job can override how it is displayed.

https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1983-jobsplugingetstatus-doesnt-display-manually-scheduled-jobs-and-displays-some-jobs-differently

### DIFF
--- a/framework/src/play/jobs/Job.java
+++ b/framework/src/play/jobs/Job.java
@@ -143,6 +143,7 @@ public class Job<V> extends Invoker.Invocation implements Callable<V> {
      */
     public void every(int seconds) {
         JobsPlugin.executor.scheduleWithFixedDelay(this, seconds, seconds, TimeUnit.SECONDS);
+        JobsPlugin.scheduledJobs.add(this);
     }
 
     // Customize Invocation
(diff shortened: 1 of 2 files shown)


xael-fry:
I am not sure about your fix for `findOrCreateJob`, as it is different than before, as you now only have one instance of the same job if using different annotation, so it may result in a different behavior

angryziber:
It may be different, but these instances are actually unexpected and
identical. Which use could you see for two instances?

xael-fry:
Some users may use a job with both @Every("24h") and @OnApplicationStart(async = true)

angryziber:
Yes, this is exactly how we use it for many jobs, and the unexpected part
was that there were 2 instances instead of one. Nothing was broken as the
instances were identical, just it was weird to see many jobs as duplicates
in application status.

I think this is a good change and it is hard to break anything with it (or
you have an idea?). But if you are still hesitating, please merge at least
other commits without this one.

xael-fry:
New PR #916

PR #911, opened 2015-09-22
* [#1946] fix(FunctionalTest): fix call of constructor StringPart

xael-fry:
* [#1941] chore(travis): test openjdk7 and oraclejdk8
* [#1941] chore(travis): remove test on openjdk6
* [#1941] chore(library): update to async-http-client from v1.8.14 to 1.9.31
* [#1946] chore(libraries): update from netty 3.9.4 to 3.10.4
** remove some deprecated method call and used the new ones

PR #912, opened 2015-09-25
[#1946] netty update master

xael-fry:


xael-fry:
Check #911 for 1.4.x branch

PR #916, opened 2015-10-07
[#1983] Fix jobs plugin

xael-fry:
…ionStart and @Every

* Previously these jobs would be displayed twice in application status.
* refactor(JobPlugin):Fixed some compiler warnings
* refactor(JobPlugin): Create executor and scheduledJobs at the same time to prevent unexpected NPEs during startup
* refactor(JobPlugin):Be consistent with how jobs are displayed in application status: use toString() in both 'Scheduled jobs' and 'Waiting jobs'
* refactor(JobPlugin):Make JobsPlugin aware of Jobs scheduled programmatically with Job.every() and display them in application status
* refactor(JobPlugin): Use Job.toString() also in monitor name, for consistency with Scheduled Jobs and Waiting Jobs

[Lighthouse #1983](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1983)


xael-fry:
Thanks

PR #922, opened 2015-11-11
Update testrunner module of 1.2.x to version of Play 1.4.0

oporkka:
Update testrunner module to get newer headless browsers than IE8 and Firefox 3.6


xael-fry:
We cannot update 1.2.x to htmlUnit 2.16, as Java 7 is required for it.
1.2.x must be compatible with Java 5.

Please upgrade to play 1.4.0

PR #924, opened 2015-11-16
disable SSLv2,SSLv3

zhaoyin:
//disable SSLv2,SSLv3  --http://googleonlinesecurity.blogspot.com/2014/10/this-poodle-bites-exploiting-ssl-30.html?m=1


xael-fry:
Will be done in #929

PR #927, opened 2015-11-20
add method Plugin.onActionInvocationFinally

asolntsev:
We need a plugin method that is guaranteed to be called after **every** action (no matter if action execution succeeded or failed). Now Play1 doesn't have such a method.

Ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2002-add-method-pluginonactioninvocationfinally


xael-fry:
Thanks

PR #928, opened 2015-11-27
* [#2003] fix(PlayHanlder): make sure current response re-set before …

xael-fry:
…parsing request, this prevent sending back response with arbitrary session


xael-fry:
Merged and pushed in master 1.3x and 1.4.x.
Thanks

xael-fry:
It was fixed in 1.2.x in #751

PR #931, opened 2015-11-30
* [#1993] test(images): fix path to run test with ant task

xael-fry:
[Lighthouse #1993](https://play.lighthouseapp.com/projects/57987/tickets/1993-imagestestjava-test-fails-when-running-ant-unittest)


xael-fry:
Thanks

PR #938, opened 2016-01-12
[#2011] "double semicolon" as escape in evolutions scripts

sbeigel:
"double semicolon" used to escape a semicolon and avoid splitting the same way Play 2.x supports escaping the single semicolon.

Useful to handle trigger or function declarations in evolutions scripts for SQL dialects that don't support dollar comments/delimiters (i.e. MySQL).

See [Lighthouse 2011](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2011-support-double-semicolon-escaping-in-evolution-scripts)


xael-fry:
Thanks

PR #941, opened 2016-02-01
[#2016] default value of "XForwardedOverwriteDomainAndPort" setting should be…

asolntsev:
… false

I believe it was initial intention of author of Pull Request #1772
See their description in https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1772: "Our change has no effects to any existing Play applications"

Lighthouse issue: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2016-default-value-of-xforwardedoverwritedomainandport-setting-should-be-false


xael-fry:
Thanks

PR #943, opened 2016-02-02
[#2017] Use StringBuilder instead of StringBuffer

sbeigel:
Use StringBuilder as a small performance boost and best-practice where possible (as a thread-safe/synchronized access is not needed).


sbeigel:
[Lighthouse issue 2017] (https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2017-use-stringbuilder-instead-of-stringbuffer)

xael-fry:
Thanks

PR #944, opened 2016-02-02
[#2018] class caches added

sbeigel:
Use caches to speed up class  & assignable classes lookup: These methods are called quite often and for a large number of classes, it's desirable to avoid iterating over all of them every time. Using a profiler, there is a clear performance boost.

### DIFF
--- a/framework/src/play/classloading/ApplicationClassloader.java
+++ b/framework/src/play/classloading/ApplicationClassloader.java
@@ -1,6 +1,8 @@
 package play.classloading;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
+
 import play.Logger;
 import play.Play;
 import play.cache.Cache;
@@ -441,10 +443,18 @@ public class ApplicationClassloader extends ClassLoader {
                     }
                 });
             }
+            allClassesByNormalizedName = new HashMap<String, ApplicationClass>(allClasses.size());
+            for (ApplicationClass clazz : Play.classes.all()) {
+                allClassesByNormalizedName.put(clazz.name.toLowerCase(), clazz);
+                if (clazz.name.contains("$")) {
+                    allClassesByNormalizedName.put(StringUtils.replace(clazz.name.toLowerCase(), "$", "."), clazz);
+                }
+            }
         }
         return allClasses;
     }
     List<Class> allClasses = null;
+    Map<String, ApplicationClass> allClassesByNormalizedName = null;
 
     /**
      * Retrieve all application classes assignable to this class.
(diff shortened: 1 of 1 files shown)


sbeigel:
[Lighthouse issue 2018](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2018-class-caches-added)

xael-fry:
Thanks

selangley:
An apparent side-effect of this change is that all static fields of ApplicationClasses are initialized before the Play plugins are initialized. One of our developers was relying on the old behavior where an ApplicationClass could be loaded (and its static fields initialized) after the plugins were initialized. 

Here are his comments:

```
I think the specific problem happened when I added a controller action to an existing controller that was not annotated with @NoTransaction.  I had been unwittingly relying on the JRE's lazy loading because some static fields needed to be initialized after play's plugins had been initialized.  When I made an unrelated change, I caused a controller class to be visible to the "class enhancer", which made it load before any plugins had been loaded.

If it's important to know the plugin that had to be initialized, I can revert my change and try to reproduce the original bug.

...

Reworked the WorkQueue to address some bootstrapping problems that caused NPEs and deadlocks.
The specific problem was the class's static initializer that created a Job that was submitted to
Play's Job queue. If the class were initialized too early in the bootstrap process, submitting the
Job would throw a NPE. The problem is that some of Play's enhancers might force early initializing
of this class in hard-to-predict ways (for example, if a controller class had an "await()" and a
static initialization of a WorkQueue.

The fix is two-fold. First, we use a Thread instead of a Job to manage the work-queue. This is
better overall, since the work queue didn't get any benefit from being a Job. This decoupled
the static initialization of SingleThreadedWorkQueue from the Jobs plugin.

The second fix is to use lazy initialization for the singleton object. This decouples the creation
of the singleton from loading of the class. This fixes a deadlock when running "play autotest".

    public static SingleThreadedWorkQueue getInstance() {
        // We must use lazy initialization (via a nested class) instead of initializing it
        // on the static variable on the outer class.  Normally, lazy initialization is done
        // to save memory on rarely used parts of the system, but in this case, we will
        // always create the instance.  The lazy initialization is to avoid a Play!
        // bootstrapping problem where a "class enhancer" causes the outer class to be loaded
        // (and therefore all of its static variables initialized) before all of the Play!
        // plugins have been initialized, leading to NullPointerExceptions and deadlocks.
        return HelperSingleton.instance;
    }

```

I suppose you couldn't suggest an easier workaround for him, or could you?

Thanks.

sbeigel:
I don't see how this change affects initial class loading. It adds caches, but the application classes were loaded before as well (see calls to getAllClasses() or Play.classes.all() or the equivalent for precompiled classes).
The first part of the change improves the lookup of the (already loaded!) app classes by normalizing their names.
The second part caches the result of assignable classes but doesn't change the class loading either.

Could you test your app with a play version w/out this PR to be sure that your problem was introduced by adding these caches?

selangley:
Yes, you are correct.  I was mistaken.  The timing of the initial class loading has nothing to do with this PR. Sorry for the confusion.

PR #946, opened 2016-02-06
[#2019] Make method Cache.clear() null-safe

asolntsev:
it's often called during Play! restart, when cache is not initialized yet

Lighthouse ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2019-make-method-cacheclear-null-safe


xael-fry:
Test failing because of travis issue

xael-fry:
Thanks

PR #947, opened 2016-02-06
[#2020] avoid multiple creation of new arrays/maps

asolntsev:
It uses excessive memory and GC resources without need.
Use Collections.emptyList() and Collections.emptyMap() instead.

Lighthouse ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2020-avoid-multiple-creation-of-new-arraysmaps


xael-fry:
Thanks

PR #951, opened 2016-02-15
fix(travis): move back to the old infrastructure

xael-fry:
Just to test travis


xael-fry:
Merged and pushed in master and 1.4.x
Thanks

PR #952, opened 2016-02-17
Add jar to be able to use ehcache as a second level cache

scesbron:
With play 1.2.x it was possible to use ehcache as hibernate second level cache. With play 1.4.1 this is not possible because there is a jar missing.

This is explained on that blog post : http://web.archive.org/web/20130117102553/http://www.javacraft.org/2012/03/migrate-to-hibernate-4-ehcache.html

With the `hibernate-ehcache jar` file it is possible to use ehcache as a second level cache by defining this entry in the `application.conf` file :

    hibernate.cache.use_second_level_cache=true
    hibernate.cache.use_query_cache=true
    hibernate.cache.region.factory_class=org.hibernate.cache.ehcache.EhCacheRegionFactory


xael-fry:
Could you stash your 3 commit in one, and put your french comment in french?
Thanks

scesbron:
I will create a new pull request from a fresh branch

xael-fry:
See #958

PR #953, opened 2016-02-17
[#2026] Blob: missing methods implemented to support (2nd level) caching

sbeigel:
Play's Blob Hibernate UserType isn't implementing all methods but throwing UnsupportedOperationExceptions for disassemble, assemble and replace. These methods are important if Hibernate's 2nd level cache is used.


sbeigel:
[Lighthouse issue #2026](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2026-blob-missing-methods-implemented-to-support-2nd-level-caching)

xael-fry:
Thanks

PR #954, opened 2016-02-21
[#2027] Cache locale

asolntsev:
Cache result of method Lang.getLocale()

this method can be called many times on every page, and it iterates all 150+ available locales.

Lighthouse ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2027-cache-method-langgetlocale


xael-fry:
Thanks

PR #955, opened 2016-02-22
* [#2015] fix(i18n): Fix i18n tag only restore to restore all '%' characters

xael-fry:
[lighthouse issue #2015](https://play.lighthouseapp.com/projects/57987/tickets/2015-i18n-tag-only-restore-the-first-character)


xael-fry:
Thanks

PR #956, opened 2016-02-22
Fix regression in WS.getString between 1.2.x and 1.4.x

scesbron:
In http-client-async getResponseBodyAsStream cannot be called multiple times. This leads to WS.getStream to also must be called once.

Here is a proposal to not use getStream in getString method so getString can be called multiple times. This was the default behaviour in play 1.2.x. The modification leads to several regressions because ofter getString is used in logs before really using the response.

With this version getString use underlying getResponseBody that can be use multiple times.


scesbron:
Close this pull-request to replace it with a clean one #959

PR #960, opened 2016-02-29
use correct encoding in method WSAsync.getString()

asolntsev:
by default `NettyResponse` uses `ISO 8859-1`, but we need `UTF-8` in most cases.

Ticket: [#2037](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2037-use-correct-encoding-in-method-wsasyncgetstring)


xael-fry:
Merged and pushed in master, 1.4.x and 1.3.x.
Thanks

PR #963, opened 2016-03-10
Catch exception that has been thrown in method annotated with @Before

braman:
```
public class SecurityCtl extends Controller {

    @Util
    @Before(priority = 0)
    public static void checkAuth() {
        //if user is authorized then ok

        if (request.cookies.containsKey("AUTH_TOKEN")) {


            final Cookie cookie1 = request.cookies.get("AUTH_TOKEN");

            final String token = cookie1.value;

            //if token exists in DB

        } else {
            throw new UnauthorizedException();
        }
    }

    @Util
    @Catch(priority = 0, value = Throwable.class)
    protected static void handleExceptions(Throwable t) {
            //if not web socket
        if (response != null) {
            response.status = 500;
            Logger.error(t, "ERROR");
            renderJSON(String.format("{\"errorCode\":%d, \"description\":\"%s\"}", response.status, t.getMessage()));
        }
    }

}
```

Controller class:

```
@With(SecurityCtl.class)
public class MyCtl extends Controller {

    public static void getUser(Long id) {
        User u = User.findById(id);
        renderJSON(u);
    }

}
```

### DIFF
--- a/framework/src/play/mvc/ActionInvoker.java
+++ b/framework/src/play/mvc/ActionInvoker.java
@@ -140,8 +140,31 @@ public class ActionInvoker {
             // 3. Invoke the action
             try {
                 // @Before
-                handleBefores(request);
+                try {
+					handleBefores(request);
+				} catch (InvocationTargetException ex) {
+					
+					// @Catch
+					Object[] args = new Object[] { ex.getTargetException() };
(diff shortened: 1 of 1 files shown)


asolntsev:
@braman We found that this PR caused a bug. See PR https://github.com/playframework/play1/pull/1048

PR #964, opened 2016-03-11
[#2041] enable manuallyconfigured Logger with windows

flybyray:
This fix should enable you to build the Play! framework entirely with all tests on windows.
Perhaps you can use this as starter (ant, jdk ... required)

`SET "JAVA_HOME=c:\Program Files\Java\jdk1.8.0_72"`
`SET "ANT_HOME=c:\apache\ant\1.9.6"`
`SET "PLAY_HOME=c:\Users\user\git\play"`
`SET "PATH=%PATH%;c:\Users\user\git\play\python;%JAVA_HOME%\bin;%ANT_HOME%\bin;%PLAY_HOME%"`
`SET "JAVA_OPTS=-Duser.language=en -Duser.region=US"`
`CD %PLAY_HOME%`
`ant -f framework\build.xml -Dversion=1.4.2 test`
`pause`

### DIFF
--- a/framework/build.xml
+++ b/framework/build.xml
@@ -247,6 +247,12 @@
     <!-- Launch all *Test.java -->
 
     <target name="test" depends="clean,jar,build-test-modules" description="run tests suite">
+        <condition property="pythonExecutable" value="${basedir}/../python/python.exe">
+            <and><os family="windows"/></and>
+        </condition>
+        <condition property="pythonExecutable" value="python">
+            <and><os family="unix"/></and>
+        </condition>
         <condition property="playExtension" value=".bat">
             <and><os family="windows"/></and>
         </condition>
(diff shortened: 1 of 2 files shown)


flybyray:
here for example a windows 7 with german locale schema
http://www.screencast.com/t/tsGVJkue

framework/build.xml
```diff
@@ -247,6 +247,12 @@
     <!-- Launch all *Test.java -->
 
     <target name="test" depends="clean,jar,build-test-modules" description="run tests suite">
+        <condition property="pythonExecutable" value="${basedir}/../python/python.exe">
+            <and><os family="windows"/></and>
+        </condition>
+        <condition property="pythonExecutable" value="python">
+            <and><os family="unix"/></and>
```
xael-fry: <os family="mac"/> are not handle. I think you should use 
<not>
        <os family="windows"/>
</not>

flybyray:
i do not know or test anything with mac(osx). but if you are interested in then check: https://ant.apache.org/manual/Tasks/condition.html
```
<condition property="isMacOsButNotMacOsX">
    <and>
      <os family="mac"/>

      <not>
        <os family="unix"/>

      </not>
    </and>
  </condition>
```
hence mac osx is part of the os family unix in that condition.
but maybe apple is doing all things their way and python is not accessible with this setting then you need to check yourself. i cannot test it.

flybyray:
closed in favor https://github.com/playframework/play1/pull/1019

PR #965, opened 2016-03-13
Upgrade dependencies

asolntsev:
just upgrade all available jar versions (except JPA/Hibernate)

Ticket: https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2042-upgrade-dependencies


xael-fry:
Same problem here, I didn't have time to make test passed for 1.4.2. That why I delay update of async-http-client / netty to 1.4.3. :)

asolntsev:
@xael-fry No problems. If I revert just netty version, is it ok?

xael-fry:
Or you can try to fix it?  ;)

asolntsev:
@xael-fry It's not simple. I tried, but did not succeed :(

I updated the pull request: reverted netty version back to 3.10.4. But some tests still fail. It seems that the problem is not caused by netty 3.10.5, but something else...

xael-fry:
Yes it is not simple, I don't really understand what is the problem, in my previous PR I revert netty and async-http-client.
So maybe the problem is from async-http-client 1.9.33.

xael-fry:
Will be handle in  #971, remove async-http to make test work

asolntsev:
@xael-fry Thank you!

PR #976, opened 2016-05-31
[#2052] make session put with Object work the same way as with String

cbxp:
https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2052-sessionput-works-incorrectly-with-null-objects

### DIFF
--- a/framework/src/play/mvc/Scope.java
+++ b/framework/src/play/mvc/Scope.java
@@ -1,13 +1,5 @@
 package play.mvc;
 
-import java.lang.annotation.Annotation;
-import java.net.URLEncoder;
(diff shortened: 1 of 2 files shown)


asolntsev:
NB! This is backward-incompatible change.
Every time when you put `null` to session, the previous code put `"null"` string:

```java
session.put("hello", (Object)null);
assertEquals("null", session.get("hello"));
```

The new code removes value from session in this case:
```java
session.put("hello", (Object)null);
assertNull(session.get("hello"));
```

xael-fry:
Merged in master and 1.4.x
Thanks

PR #977, opened 2016-05-31
[#2051] do not allow direct execution of @Util and @Catch methods

cbxp:
Fix for

https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2051-it-is-possible-to-call-util-and-catch-controller-methods-directly-using-a-http-request


framework/src/play/mvc/ActionInvoker.java
```diff
@@ -593,8 +612,8 @@ static Object invokeWithContinuation(Method method, Object instance, Object[] re
                             new Exception("class " + controller + " does not extend play.mvc.Controller"));
                 }
             }
-            actionMethod = Java.findActionMethod(action, controllerClass);
-            if (actionMethod == null) {
+            actionMethod = findActionMethod(action, controllerClass);
+            if (actionMethod == null || !isActionMethod(actionMethod)) {
```
xael-fry: I don't think we need to test ```isActionMethod``` again as you already test in in your method ```findActionMethod```.

xael-fry:
Merged and pushed in master.
Thanks

Issue #985, opened 2016-06-27
Play should NOT clear cache after every code change


asolntsev:
Now Play clears cache after every code change.

See `play.classloading.ApplicationClassloader:333`:

``` java
public void detectChanges() {
        ...
        if (!newDefinitions.isEmpty()) {
            Cache.clear();
        ...
}
```

It was done in this commit:

```
4f8789d Jean-Francois Poux on 18/09/09 at 12:47
Clear cache when hotswapping code
```

I believe it's a bad idea, because cache may contain lot of data that is hard to load again. No need to lose this data.

theo:
I believe we want to clear because Cache relies on java Serializable. Class changes can easily break java serialization resulting in entries in the cache that the application is unable to deserialize.

asolntsev:
@theo Yes, this reason is clear. Sometimes cache can contain classes that have been modified. But in happens rarely. In most cases, cache still contains valid objects.

At the same time, the reason "cache may contain lot of data that is hard to load again" is still very important for us. In our project this is a problem.

So I suggest to solve this problem in another way. Let's leave cache untouched, but add try/catch to method `Cache.get()`. This method can return null if it failed to deserialize object which has been recently changed.

asolntsev:
Yes,
we also experience the same problem.
When we install new release, we get lot of warnings in logs (we have a
wrapper for Cache with try/catch inside).
Yes, I hope we can solve this problem in that way.

Andrei Solntsev

2016-06-30 21:08 GMT+03:00 Fraserhardy <notifications@github.com>:

asolntsev:
@theo @Fraserhardy @xael-fry 
Please review the pull request #999 that should fix this problem.

Proposal is simple: Play does not clear cache on reload, but ignores all deserialization errors during 1 minute after restart. This period is configurable via `play.cache.warmupPeriodMs` property (default value is 60000).

theo:
It seems bad practice to swallow exceptions, especially during startup.

This seems like something the client needs to handle.  If a client attempts to pull something from cache and is not able to deserialize, the client is in the best position to handle the error (ignore or some other remediation)

asolntsev:
@theo I agree with you, I also don't like swallowing exceptions. 
But in this case, Play swallows exceptions only during 1 minute. If you have a problem, you will start getting it after a minute. You will not miss it.

Anyway, I would like to implement a better solution, but I failed to do it.
My ideal algorithm would be:
1. Get object from cache
[...]

PR #999, opened 2016-08-31
Do not clear cache on reload

asolntsev:
This pull request fixes the issue #985 

Proposal is simple: Play does not clear cache on reload, but ignores all deserialization errors during 1 minute after restart. This period is configurable via `play.cache.warmupPeriodMs` property (default value is 60000).


asolntsev:
Ok, I agree that this is not the best possible solution. Let's cancel the pull request. I will implement another one.

PR #1000, opened 2016-09-06
[#2062] Reverse routing improvement

vadimv82:
Here is simplest solution for the issue #2062


vadimv82:
There is some problem with build. Failure can not be related to changes.
*** buffer overflow detected ***: /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java terminated

Issue #1009, opened 2016-10-14
Do not log error and exception when compiling a temaplate with a non GroovyInlineTag


xael-fry:
Report by @selangley  in #933

PR #1010, opened 2016-10-14
[#1009] fix(templates): not log error and exception when compilin a template with a non GroovyInlineTag

xael-fry:
it revert modif on GroovyTemplateCompiler.java don in 765dec207a563d80069665d0df40cc32c6ffec6a


xael-fry:
fox issue #1009

PR #1012, opened 2016-10-18
[#2064] Multilangual routes files support

vadimv82:
Multiple multilangual routes files support.
You can put several routes files into conf folder in a way:
routes.en, routes.fr, routes.ru.
Same route can have multiple urls for different languages.
Useful for SEO.

### DIFF
--- a/framework/src/play/Play.java
+++ b/framework/src/play/Play.java
@@ -114,7 +114,7 @@ public class Play {
     /**
      * Main routes file
      */
-    public static VirtualFile routes;
+    public static List<VirtualFile> routes;
     /**
      * Plugin routes files
      */
@@ -181,6 +181,11 @@ public class Play {
      * as a WAR in an applicationServer
      */
     public static boolean standalonePlayServer = true;
+
+    /**
+     * This flag indicates that app has multiple routes files for different locales
+     */
+    public static boolean multilangRouteFiles = false;
 
     /**
      * Init the framework
@@ -276,7 +281,7 @@ public class Play {
         }
 
         // Main route file
-        routes = appRoot.child("conf/routes");
+        routes = loadRoutesFiles(appRoot);
 
         // Plugin route files
         modulesRoutes = new HashMap<>(16);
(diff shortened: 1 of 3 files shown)


vadimv82:
I added tests, and changed some logic.

Now you can add specific language routes file. For example routes.ru_RU, routes.fr_FR.
Those can contain language specific urls. If you open lang specific url it will reverse right action, but also automatically activates language as it is specified in Route, by url.
All that is not lang specific, like ajax calls, static content servers e.t.c you can put under "routes" file without lang extension. All those calls will not activate language change.

vadimv82:
Can please somebody review my pull request?

xael-fry:
@vadimv82 Could you stash your changes?

framework/src/play/mvc/Router.java
```diff
@@ -614,6 +622,26 @@ public static ActionDefinition reverse(String action, Map<String, Object> args)
...
+  /**
+   * Prioritize action routes based on active locale and Play.langs properties. Active lang has highest priority, then prioritized according to Play.langs order.
+   *
+   * @param matchingRoutes
+   */
+  private static void prioritizeActionRoutesBasedOnActiveLocale(List<Router.ActionRoute> matchingRoutes) {
+        if(matchingRoutes.size()==0) return;
+        final String locale = Lang.get();
+        if(StringUtils.isEmpty(locale)) return;
+        matchingRoutes.sort(new Comparator<ActionRoute>() {
```
xael-fry: Jenkins cannot compile the code 
```
    [javac]         matchingRoutes.sort(new Comparator<ActionRoute>() {
    [javac]                       ^
    [javac]   symbol:   method sort(<anonymous Comparator<ActionRoute>>)
    [javac]   location: variable matchingRoutes of type List<ActionRoute>
```
Seems like some method are missing as you use the interface Comparator

framework/src/play/Play.java
```diff
@@ -114,7 +114,7 @@ public boolean isProd() {
     /**
      * Main routes file
      */
-    public static VirtualFile routes;
+    public static List<VirtualFile> routes;
```
asolntsev: It breaks backward compatibility.
I suggest leaving `VirtualFile routes` untouched and adding a new field `List<VirtualFile> internationalizedRoutes`. In this case you don't need field `multilangRouteFiles`.

framework/src/play/mvc/Router.java
vadimv82: I run same build on my machine with ant -buildfile ./framework/build.xml test
And build is succesfull.

I use anonymous comparator

`sortedMatchingRoutes.sort(new Comparator<ActionRoute>() {
            @Override
            public int compare(ActionRoute ar1, ActionRoute ar2) {
                if(locale.equals(ar1.route.locale)) return -1;
                if(locale.equals(ar2.route.locale)) return 1;
                return Integer.compare(Play.langs.indexOf(ar1.route.locale), Play.langs.indexOf(ar2.route.locale));
            }
        });`

PR #1019, opened 2016-10-30
[#2067] fix os and language related stuff

flybyray:
- set language for test execution else MessagesTest will fail on non en_US systems
- fix wrong filepath assumption (based on string comparison) for Logger.configuredManually flag
- fix assertEquals assumptions of FastTagsTest for os independant line.separator
- fix python script for windows

now just run the test target of the ant build file: http://www.screencast.com/t/JwT1f6dj


flybyray:
https://travis-ci.org/playframework/play1/builds/171810044#L83
"Apache Ant(TM) version 1.8.2 compiled on December 3 2011"

Why is travis using an old ant version?

flybyray:
outdated: https://play.lighthouseapp.com/projects/57987/tickets/2067-fix-os-and-language-related-stuff
=> splitting into several pull requests
#1147 
#1148 
...

PR #1020, opened 2016-10-31
[#1919] fixes Building modules with a space in path under windows fails

flybyray:
https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1919-building-modules-with-a-space-in-path-under-windows-fails


asolntsev:
@flybyray thank you! merged.

PR #1023, opened 2016-11-12
Check handle multiple value

pengzhang:


asolntsev:
@pengzhang  Thank you for the pull request!
Can you describe the case? What problem are you trying to fix?

PR #1025, opened 2016-11-13
Detect changed for JNDI datasource which using jndi: prefix

phongphan:
The jndi: prefix was added at https://github.com/playframework/play1/pull/441.
But the `changed()` might not be detected when calling `onApplicationStart` in DBPlugin.java (it will return `false` if db.url and db.driver is not configured). And the datasource with `jndi:` prefix will not be initialized and not been added to DB.datasources.


phongphan:
This issue also exists in 1.2.x, 1.3.x, and play1 master. Do you have plan to backport this fixed? Or Could I open separate PRs to each branch (I am more than happy to do it)?

Since there are not much Glassfish users. And I only use 1.4

asolntsev:
Hi @phongphan !
It's easier for me to merge 3 separate PRs. Thank you in advance.

PR #1032, opened 2016-11-26
Disable generating accessors

asolntsev:
add possibility to disable generating getters/setters 
(via setting `play.propertiesEnhancer.generateAccessors=false`)

it's useful because generation of getters/setters and replacing all field accesses with getters/setters is the slowest part of compilation process. Many applications do not really need it.


asolntsev:
it was bad PR. I will create another one.

PR #1033, opened 2016-11-27
add possibility to disable generating getters/setters

asolntsev:
add possibility to disable generating getters/setters 
(via setting `play.propertiesEnhancer.generateAccessors=false`)

it's useful because generation of getters/setters and replacing all field accesses with getters/setters is the slowest part of compilation process. Many applications do not really need it.


xael-fry:
@asolntsev Could you update the documentation for that new parameter?

asolntsev:
@xael-fry I have updated documentation in branch 1.4.x. Thank you for the reminder!

Issue #1031, opened 2016-11-24
transitiveDependencies in dependencies.conf doesn't work


jadeglaze:
Hi All,

I've discovered that the transitiveDependencies global property in dependencies.conf can never be successfully set to false.

In YamlParser.java line 115 we have:

```java
boolean transitiveDependencies = get(data, "transitiveDependencies", boolean.class, true);
```

which uses this pair of functions at YamlParser.java 278:

```java
    <T> T get(Map data, String key, Class<T> type) {
        if (data.containsKey(key)) {
            Object o = data.get(key);
            if (type.isAssignableFrom(o.getClass())) {
                if(o instanceof String) {
                    o = o.toString().replace("${play.path}", System.getProperty("play.path"));
                    o = o.toString().replace("${application.path}", System.getProperty("application.path"));
                }
                return (T) o;
            }
        }
        return null;
    }

    <T> T get(Map data, String key, Class<T> type, T defaultValue) {
        T o = get(data, key, type);
        if (o == null) {
            return defaultValue;
        }
        return o;
    }
```

The problem is the use of `boolean.class` instead of `Boolean.class` on line 115 combined with the check `type.isAssignableFrom(o.getClass())` in the get function. `boolean.class.isAssignableFrom(Boolean.class)` is `false`.

The fix on line 115 is simple (use `Boolean.class`):
```java
boolean transitiveDependencies = get(data, "transitiveDependencies", Boolean.class, true);
```

Note that there doesn't appear to be any unit test coverage of transitiveDependencies.
(I'm also suspicious that the exclude suboption in the require section may not work - and it has no unit test coverage - but I'm still debugging that.)

Thanks!

PR #1034, opened 2016-11-28
[#1031] - fix transitiveDependencies global setting in dependency.yml

jadeglaze:


jadeglaze:
Hmm. I think it's highly unlikely that my change broke the build. Thoughts?

PR #1039, opened 2016-12-14
avoid thousands of MissingPropertyExceptions

angryziber:
Binding can only contain variables (basically, a Map), not properties.

If groovy doesn't find a variable, it throws a MissingPropertyException populating the stacktrace.

GroovyTemplate.getProperty() first calls the getVariable(), if it throws (happens very often if you have optional bindings), then it catches and delegates to getProperty(), that does some slow work itself again.

This small change avoids these excessive exceptions with stacktraces on every missing variable access.


xael-fry:
We revert this PR for now as I found some problems, I created #1067 and will add some tests before review this PR again

PR #1042, opened 2016-12-14
[#2071] precompile additional templates

flybyray:
see https://play.lighthouseapp.com/projects/57987/tickets/2071

### DIFF
--- a/framework/src/play/templates/TemplateLoader.java
+++ b/framework/src/play/templates/TemplateLoader.java
@@ -222,18 +222,18 @@ public class TemplateLoader {
             }
         }
 
-        String play_compile_templates = Play.configuration.getProperty(
-                "play_compile_templates",
+        String play_templates_compile = Play.configuration.getProperty(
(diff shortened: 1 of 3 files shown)


framework/src/play/templates/TemplateLoader.java
```diff
@@ -217,6 +221,46 @@ public static Template load(String path) {
...
             }
         }
+
+        String play_compile_templates = Play.configuration.getProperty(
+                "play_compile_templates",
+                System.getProperty(
+                        "play_compile_templates",
+                        System.getenv("PLAY_COMPILE_TEMPLATES")));
+        String play_compile_templates_path_separator = Play.configuration.getProperty(
+                "play_compile_templates_path_separator",
```
xael-fry: We used '.' to separate word in config param name so should be @play.compile.templates.path.separator@

flybyray: thanks for the hints. pushed an update.

PR #1049, opened 2016-12-21
Make writing pure unit tests easier

angryziber:
Initialize most common Play.* fields to sane default values.

Plus fix duplication and remove some dead code.


framework/src/play/i18n/Messages.java
```diff
@@ -32,9 +32,9 @@
 
     private static final Object[] NO_ARGS = new Object[] { null };
 
-    static public Properties defaults;
+    public static Properties defaults = new Properties();
 
-    static public Map<String, Properties> locales = new HashMap<>();
+    public static Map<String, Properties> locales = new HashMap<>();
```
theo: If removing the null checks below , make ```locales```  and ```defaults ``` final?

asolntsev: @theo 
* `defaults` is re-initialized in `MessagesPlugin` and many tests - so it cannot be final. 
* I did `locales` final in branch 1.4.x

Issue #1051, opened 2016-12-22
Play dependencies silently resolve not found dependencies


xael-fry:
On local you a module "test" version1.16.
It deps if you set "test"  1.15, the dependencies will not resolve the module but no error is displayed.
If you use verbose parameters, you will have:
```
pxpLocalModules: bad revision found in E:\play-modules\test\conf\dependencies.yml: expected='1.16 found='1.16'
````
But no error in the console

xael-fry:
Merged in 1.4.x

PR #1054, opened 2016-12-23
* [#1051] fix(DependenciesManager): display error when running dependencies

xael-fry:
cf #1051

### DIFF
--- a/framework/src/play/deps/DependenciesManager.java
+++ b/framework/src/play/deps/DependenciesManager.java
@@ -4,13 +4,13 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.ivy.Ivy;
+import org.apache.ivy.core.module.id.ModuleRevisionId;
 import org.apache.ivy.core.report.ArtifactDownloadReport;
 import org.apache.ivy.core.report.ResolveReport;
 import org.apache.ivy.core.resolve.IvyNode;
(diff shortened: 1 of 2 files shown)


framework/src/play/deps/DependenciesManager.java
```diff
@@ -183,36 +203,52 @@ public boolean problems() {
...
-        if (!missing.isEmpty()) {
+
+        if (!missing.isEmpty() || !problems.isEmpty()) {
             System.out.println("~");
             System.out.println("~ WARNING: Some dependencies could not be downloaded (use --verbose for details),");
             System.out.println("~");
             for (ArtifactDownloadReport d : missing) {
-                String msg = d.getArtifact().getModuleRevisionId().getOrganisation() + "->" + d.getArtifact().getModuleRevisionId().getName() + " " + d.getArtifact().getModuleRevisionId().getRevision() + ": " + d.getDownloadDetails();
-                System.out.println("~\t" + msg);
+                StringBuilder msg = new StringBuilder(d.getArtifact().getModuleRevisionId().getOrganisation()).append(" -> ")
```
asolntsev: We actually don't need StringBuilder here. Concatenating String with `+` is more readable, and works the same.

framework/src/play/deps/DependenciesManager.java
```diff
@@ -318,58 +354,55 @@ private boolean isPlayModule(ArtifactDownloadReport artifact) throws Exception {
...
 
         return resolveEngine.resolve(ivyModule.toURI().toURL(), resolveOptions);
     }
 
     public Ivy configure() throws Exception {
-
         boolean verbose = System.getProperty("verbose") != null;
         boolean debug = System.getProperty("debug") != null;
         HumanReadyLogger humanReadyLogger = new HumanReadyLogger();
+        this.logger = humanReadyLogger;
```
asolntsev: Let's remove local variable `humanReadyLogger` and use `logger` instead.

framework/src/play/deps/DependenciesManager.java
```diff
@@ -26,18 +26,38 @@
 import play.libs.IO;
 
 public class DependenciesManager {
-   
+
     public static void main(String[] args) throws Exception {
+        String applicationPath = System.getProperty("application.path");
```
xael-fry: It will avoid exception if we used incorrectly the method, this way we know that all this variable are required

Issue #1064, opened 2016-12-26
Initialize play static field early


asolntsev:
... to avoid `NullPointerException`s and make writing plain unit-tests easier.

Problem description: 
* Play has lot of `public static` field - first of all, `Play.configuration`
* Many classes do use these `public static` fields
* When writing unit-test or main method, developers always need to write null checks and boilerplate code for Play initialization, e.g.: 
`if (Play.configuration == null) Play.coniguration = new Properties();`

We want to improve it.
Let's initialize **early** as many Play fields as possible.

xael-fry:
@asolntsev .
Thanks, merged and pushed in 1.4.x

PR #1065, opened 2016-12-26
#1064 add default value for static fields to make it easier to write …

asolntsev:
…pure unit tests:

* Play.configuration
* Play.id
* Play.javaPath
* Play.applicationPath
* Play.mode
* Play.templatesPath
* Play.modulesRoutes
* Messages.defaults

(PR for issue https://github.com/playframework/play1/issues/1064)

### DIFF
--- a/framework/src/play/i18n/Messages.java
+++ b/framework/src/play/i18n/Messages.java
@@ -32,11 +32,11 @@ public class Messages {
 
     private static final Object[] NO_ARGS = new Object[] { null };
 
-    static public Properties defaults;
+    public static Properties defaults = new Properties();
 
-    static public Map<String, Properties> locales = new HashMap<>();
+    public static Map<String, Properties> locales = new HashMap<>();
 
-    static Pattern recursive = Pattern.compile("&\\{(.*?)\\}");
+    private static final Pattern recursive = Pattern.compile("&\\{(.*?)\\}");
 
     /**
      * Given a message code, translate it using current locale. If there is no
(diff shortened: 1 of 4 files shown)


framework/src/play/i18n/Messages.java
```diff
@@ -107,7 +107,7 @@ public static String getMessage(String locale, Object key, Object... args) {
         if (value == null && locale != null && locale.length() == 5 && locales.containsKey(locale.substring(0, 2))) {
             value = locales.get(locale.substring(0, 2)).getProperty(key.toString());
         }
-        if (value == null && defaults != null) {
+        if (value == null) {
```
xael-fry: I think you should keep the test `default != null` as `defaults ` is public it can be set to null outside the class

framework/src/play/i18n/Messages.java
```diff
@@ -196,7 +196,7 @@ public static Properties all(String locale) {
             return defaults;
         Properties mergedMessages = new Properties();
         mergedMessages.putAll(defaults);
-        if (locale != null && locale.length() == 5 && locales.containsKey(locale.substring(0, 2))) {
+        if (locale.length() == 5 && locales.containsKey(locale.substring(0, 2))) {
```
xael-fry: I think you should keep the test `locale != null` as locale is public it can be set to null outside the class

framework/src/play/i18n/Messages.java
```diff
@@ -196,7 +196,7 @@ public static Properties all(String locale) {
...
             mergedMessages.putAll(locales.get(locale.substring(0, 2)));
         }
         mergedMessages.putAll(locales.get(locale));
```
xael-fry: We should as test on `locale` before using it

Issue #993, opened 2016-07-12
Do not swallow cause of exception


asolntsev:
We can see the following construction in many places in Play 1 framework:

```java
try {
 ...
}
catch (Exception ex) {
  throw new UnexpectedError("Could not do something");
}
```

Such code is pretty bad because the original exception `ex` gets lost. 

We should always re-throw the original exception:

```java
throw new UnexpectedError("Could not do something", ex);
```

or at least log it.

asolntsev:
Done in branch 1.4.x for many cases (but unfortunately not for all).

selangley:
While it is a good philosophy to not swallow exceptions, the code now no longer executes the original author's intent.

The code is meant to identify whether the current start tag is a GroovyInlineTag, i.e., an #{if...} #{else...} or #{list...}

If it is not, then the code should swallow the exception and continue to the next section to note other characteristics of the tag.
 
You should probably put back the original exception handler of:

// do nothing here

And add more code comments so that future maintainers have a clearer understanding of what's happening in this method.

asolntsev:
@selangley I agree, I enabled too much logging. I am going to remove some of them.

xael-fry:
@asolntsev It is the problem I talked you about, try to keep your change and avoiding the exception, but cannot find a way to do it for now

Issue #983, opened 2016-06-26
`--sync` should be a default option for `play deps`


asolntsev:
I believe that the command `play deps` without `--sync` option is absolutely useless.
You always need to delete old dependencies when you are downloading new dependencies.
Otherwise you will have duplicate dependencies in the application.

In our projects, we always write `play deps --sync` which is just boring. 

I suggest that `--sync` should be ON by default.

tazmaniax:
Agree

asolntsev:
@xael-fry It seems that everyone is agree. What do you think?

xael-fry:
Ok we can do it, and maybe add a ```--nosync``` option, so all users will be able to choose the behavior they want

xael-fry:
Merged in 1.4.x with PR 1076.
Thanks

daohodac:
I used to put my external jars in libs. Now, they are scrubbed when I run `play dependencies` Where should I put them now?

xael-fry:
@daohodac  you can use ```--nosync``` to avoid removing the lib, on you can add your lib in dependencies.yml

asolntsev:
Yes, I also think that adding your lib in dependencies.yml is a right
option.

Andrei Solntsev

2017-03-13 12:08 GMT+02:00 Alexandre Chatiron <notifications@github.com>:

daohodac:
How can I add the lib in dependencies.yml?

I assume that I have a folder in my play application, say "custom-deps" with several jars. How can I add this ref in the dependencies.yml?

thank's!

asolntsev:
@daohodac See https://github.com/codeborne/play-pdf/ for example.

You need to declate in dependencies.yml:

```
require:
    - local-jar -> flying-saucer-core-cbpatched latest.integration
    - local-jar -> flying-saucer-pdf-cbpatched latest.integration

repositories:
    - local-jars:
        type:       local
        artifact:   ${application.path}/jars/[module].jar
        contains:
            - local-jar -> *
```

daohodac:
Thank's Andrei, it worked like a charm. and now I stick to the best practice

Dao Hodac

PR #1076, opened 2016-12-30
[#983] feat(DependenciesManager): sync option should be ON by default.

xael-fry:
[#983] feat(DependenciesManager): add new option "nosync"
for issue #993

### DIFF
--- a/framework/src/play/deps/DependenciesManager.java
+++ b/framework/src/play/deps/DependenciesManager.java
@@ -135,7 +135,7 @@ public class DependenciesManager {
             }
         }
 
-        boolean autoSync = System.getProperty("sync") != null;
+        boolean autoSync = System.getProperty("nosync") == null;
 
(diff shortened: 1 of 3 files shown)


framework/src/play/deps/DependenciesManager.java
```diff
@@ -150,7 +150,7 @@ public void sync(List<File> installed) {
             System.out.println("~");
             System.out.println("~ *****************************************************************************");
             System.out.println(
-                    "~ WARNING: Your lib/ and modules/ directories are not synced with current dependencies (use --sync to automatically delete them)");
+                    "~ WARNING: Your lib/ and modules/ directories are not synced with current dependencies (don't used --nosync to automatically delete them)");
```
tazmaniax: Remove unnecessary 'd' so should be "(don't use --nosync to automatically delete them)

PR #1078, opened 2017-01-03
make method Controller.renderTemplate(...Map) public

asolntsev:
sometimes we want to render template, passing a map of arguments to it


asolntsev:
sorry, I cancel this PR. It's not needed.

PR #1107, opened 2017-02-13
Catch top-level exceptions and log them properly

cbxp:
Many crashes during start in PROD mode are not logged anywhere and are difficult to analyze.
For example this will log invalid cron exceptions properly when running in PROD mode.


asolntsev:
@xael-fry I have reviewed the pull request. It's very small and simple, recommended to merge.

(except that commit message doesn't contain issue id, but actually I don't think it's really needed. Pull request is issue in essence.)

PR #1110, opened 2017-02-20
[#1112] Allow to limit the size of whole request or the size of one uploaded file in a request

cbxp:
…file in a request.

 The configuration parameters are: 'upload.sizeMax' and 'upload.fileSizeMax'.
 The defaults stay the same -1 which means 'unlimited'.

### DIFF
--- a/framework/src/play/data/parsing/ApacheMultipartParser.java
+++ b/framework/src/play/data/parsing/ApacheMultipartParser.java
@@ -610,14 +610,14 @@ public class ApacheMultipartParser extends DataParser {
     // ----------------------------------------------------------- Data members
     /**
      * The maximum size permitted for the complete request, as opposed to
-     * {@link #fileSizeMax}. A value of -1 indicates no maximum.
+     * {@link #maxFileSize}. A value of -1 indicates no maximum.
      */
-    private long sizeMax = -1;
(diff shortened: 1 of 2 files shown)


asolntsev:
@xael-fry This is definitely good PR. I suggest to merge it ASAP.

cbxp:
@xael-fry We have updated the documentation.
By the way, I suggest to put some reasonable default values. Say, 
```
upload.maxFileSize=10240   (10 mb)
upload.maxRequestSize=30720  (30 mb)
```

Without that, it's possible that hacker can upload 100TB file to server.

framework/src/play/data/parsing/ApacheMultipartParser.java
```diff
@@ -612,12 +612,12 @@ public void setHeaders(FileItemHeaders pHeaders) {
      * The maximum size permitted for the complete request, as opposed to
      * {@link #fileSizeMax}. A value of -1 indicates no maximum.
```
xael-fry: fileSizeMax no longer exist as you rename it

asolntsev:
@xael-fry Fixed parameter name in javadoc. What about setting default values to something reasonable?

xael-fry:
@asolntsev I don't think we should define default value here, some user handle request size limit with a a front apache server, so setting new value may cause conflict when upgrading

PR #1120, opened 2017-03-09
[#2079] fix(Files): Change the method to compare if two file refers to the same one

xael-fry:
[Lighthouse #2079](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2079)

### DIFF
--- a/framework/src/play/Play.java
+++ b/framework/src/play/Play.java
@@ -726,7 +726,8 @@ public class Play {
         if (System.getenv("MODULES") != null) {
             // Modules path is prepended with a env property
             if (System.getenv("MODULES") != null && System.getenv("MODULES").trim().length() > 0) {
-                for (String m : System.getenv("MODULES").split(System.getProperty("os.name").startsWith("Windows") ? ";" : ":")) {
+
+                for (String m : System.getenv("MODULES").split(File.pathSeparator)) {
                     File modulePath = new File(m);
                     if (!modulePath.exists() || !modulePath.isDirectory()) {
                         Logger.error("Module %s will not be loaded because %s does not exist", modulePath.getName(),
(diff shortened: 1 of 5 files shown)


framework/src/play/utils/Utils.java
```diff
@@ -169,8 +169,7 @@ public static SimpleDateFormat getHttpDateFormatter() {
     }
 
     public static void kill(String pid) throws Exception {
-        String os = System.getProperty("os.name");
-        String command = (os.startsWith("Windows")) ? "taskkill /F /PID " + pid : "kill " + pid;
+        String command = (Os.isWindows()) ? "taskkill /F /PID " + pid : "kill " + pid;
```
asolntsev: remove extra braces around `Os.isWindows()`

framework/src/play/utils/Os.java
```diff
@@ -0,0 +1,27 @@
+package play.utils;
+
+/**
+ * Os detections
+ */
+public class Os {
```
asolntsev: What about renaming `Os` to `OS`?
It seems more natural for me because "OS" is a well-known acronym.

framework/src/play/Play.java
```diff
@@ -726,7 +727,7 @@ public static void loadModules(VirtualFile appRoot) {
         if (System.getenv("MODULES") != null) {
             // Modules path is prepended with a env property
             if (System.getenv("MODULES") != null && System.getenv("MODULES").trim().length() > 0) {
-                for (String m : System.getenv("MODULES").split(System.getProperty("os.name").startsWith("Windows") ? ";" : ":")) {
+                for (String m : System.getenv("MODULES").split(Os.isWindows() ? ";" : ":")) {
```
asolntsev: Can't we use just `File.pathSeparator` instead of this tricky hack? 
According to javadoc,
> On UNIX systems, this character is <code>':'</code>; on Microsoft Windows systems it is <code>';'</code>.

xael-fry: Create another issue #1121

PR #1129, opened 2017-04-10
c3p0 connection status

zhaoyin:
add c3p0 connection status index for Display the running application's status


asolntsev:
@zhaoyin Thank you for the PR. Going to include it in play 1.5 release. Sorry for the delay.

Issue #1113, opened 2017-02-23
Upgrade Hibernate to 5.x


asolntsev:
Currently Play1 uses Hibernate 4.2.19 which is pretty old.
Need to upgrade to Hibernate 5.2.5 (or later). 

See https://play.lighthouseapp.com/projects/57987-play-framework/tickets/1818-hibernate-need-to-be-updated-to-v435#ticket-1818-6

tomparle:
Thank you Andrei for this (among several others) improvement !

asolntsev:
@tomparle @xael-fry 
Hi! Can you help me with Hibernate 5?

Can anyone tell why the third line saves `BLUP`, thought I didn't call any `save` method?

```java
Horse existing = Horse.all().first();   // existing.bulp == null
existing.setBlup(new BLUP());
em().flush();   // why it executes "insert into BLUP (id, total) values (null, ?)" ???
```

asolntsev:
@tomparle @xael-fry
I found the answer. We just have to add the following property: `properties.put("org.hibernate.flushMode", FlushMode.MANUAL);`

to avoid auto-saving of dirty (unsaved) objects.

flybyray:
@asolntsev maybe it should be fixed in current play to?


https://github.com/playframework/play1/blob/master/framework/src/play/db/jpa/JPA.java#L364

The key-pattern for properties should honor `org.hibernate`
https://github.com/playframework/play1/blob/master/framework/src/play/db/Configuration.java#L88

thx

flybyray:
@xael-fry @asolntsev 

> I found the answer. We just have to add the following property: `properties.put("org.hibernate.flushMode", FlushMode.MANUAL);`


It should have a ticket in lighthouse and should be fixed for 1.4.x too. The current AUTO-FlushMode has unpredictable consequences.

asolntsev:
@flybyray @tomparle @xael-fry Actually I am not sure about my previous answer. Probably the default value `AUTO` of setting `"org.hibernate.flushMode"` is good. 

Alex, do you know why `HibernateInterceptor` is needed? Can we remove it? Without it, the failing test `JPASave` becomes green.

flybyray:
@asolntsev `HibernateInterceptor` is some essential part of Play! JPA improvment. I think you will need it even in Hibernate 5+. Please look for the `willBeSaved` variable ([search](https://github.com/playframework/play1/search?utf8=%E2%9C%93&q=willBeSaved&type=)) and its usage.

[...]

PR #1132, opened 2017-04-26
fix for issue mentioned in issue 1113 comment 289165754

flybyray:
see https://github.com/playframework/play1/issues/1113#issuecomment-289165754

this is for 1.3.x branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,6 @@
 language: java
 jdk:
   - openjdk6 
-before_install:
-  - cat /etc/hosts # optionally check the content *before*
-  - sudo hostname "$(hostname | cut -c1-63)"
-  - sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -c1-63)/" /etc/hosts | sudo tee /etc/hosts
-  - cat /etc/hosts # optionally check the content *after*
 script: ant -buildfile ./framework/build.xml test
 after_failure: 
     cat ./samples-and-tests/just-test-cases/test-result/*.failed.html
--- a/framework/src/play/db/jpa/JPAPlugin.java
+++ b/framework/src/play/db/jpa/JPAPlugin.java
@@ -2,6 +2,7 @@ package play.db.jpa;
 
 import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.log4j.Level;
+import org.hibernate.ejb.AvailableSettings;
 import org.hibernate.ejb.Ejb3Configuration;
 
 import play.Logger;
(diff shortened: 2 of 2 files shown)


asolntsev:
@flybyray Are you sure that `MANUAL` is better than `AUTO`? Why?
I am not sure at all.

Actually I solved the mentioned issue by patching hibernate-core.jar with [The Patch](https://github.com/playframework/play1/blob/a1f35221d63414e704d2b92200264d0738314c77/framework/patches/hibernate-4.2.19-patch-play.patch.patch) which disbled Hibernate built-in **dirty checking** mechanism.

flybyray:
1. If `AUTO` was the default, it should stay the default.
   1. but note that the switch to `MANUAL` will not have to much influence to simple applications. It will change runtime behavior, like hibernate cache usage. It can speed up database (less querries).
1. But it should be(have to be) configurable in `application.conf` and documented, similar like `db.isolation`

I can just tell from a complex application for risk calculations, workflow and reporting engine (e.g. hibernate 'triggers' on changes). `AUTO` flush will run your application in situations where you do not want to auto flush triggering changes within calculations (therefore it must be configurable!!!). even on some db it will produce locks on database level, somehow it depends on the `db.isolation`, but most off the time applications use read_commited. i am not sure about it but i might remember some cases where some optimisticlock exceptions occured because of `AUTO` flush. We 'hacked' the entity-manager session to change that behavior in specific cases. But I think for a framework like Play! it should not be the solution for it.

Personally i really believe that `MANUAL` should be the right choice for the Play!-framework. [Transaction management](https://www.playframework.com/documentation/1.4.x/jpa#anametransactionsTransactionmanagementa)
> Play will automatically manage transactions for you. It will start a transaction for each HTTP request and commit it when the HTTP response is sent.

In combination with [Keep the model stateless](https://www.playframework.com/documentation/1.4.x/model#anamestatelessKeepthemodelstatelessa) an auto-flush in hibernate would introduce a break of this: 
> Play is designed to operate in a ‘share nothing’ architecture.

At all do not forget [Explicit save](https://www.playframework.com/documentation/1.4.x/jpa#anamesaveExplicitsavea)
> All the persistent objects extending the JPASupport/JPAModel will not be saved without an explicit call to the save() method.

> If your code throws an exception, the transaction will automatically rollback.

asolntsev:
@flybyray I agree, it definitely should be configurable.
I am not sure about the default value. It can break existing applications. 

@xael-fry What do you think?

flybyray:
created https://github.com/playframework/play1/pull/1139
Hibernate will use AUTO by default. You have to override with a setting in application.conf

PR #1141, opened 2017-05-16
[#2121] Move _key() after reference/null checks in JPABase.equals()

davidcostanzo:
The pull request moves _key() and adds unit tests for JPABase.equals() to give confidence that there were no behavioral changes.

### DIFF
--- a/framework/src/play/db/jpa/JPABase.java
+++ b/framework/src/play/db/jpa/JPABase.java
@@ -274,14 +274,14 @@ public class JPABase implements Serializable, play.db.Model {
      */
     @Override
     public boolean equals(Object other) {
-        Object key = this._key();
-
         if (other == null) {
             return false;
         }
         if (this == other) {
             return true;
         }
+
+        Object key = this._key();
         if (key == null) {
             return false;
         }
--- a/samples-and-tests/just-test-cases/app/models/ArrayIdEntity.java
+++ b/samples-and-tests/just-test-cases/app/models/ArrayIdEntity.java
@@ -1,0 +1,20 @@
+package models;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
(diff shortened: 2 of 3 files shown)


asolntsev:
@davidcostanzo Can you describe what problem you are trying to solve? Why the check should be moved? Why are you confident that it will not break existing applications?

davidcostanzo:
Thanks for considering the pull request.  Answers to your questions follow:

>  Can you describe what problem you are trying to solve?

Briefly, the problem I'm trying to solve is to improve performance of my application.  The problem is described abstractly in [the lighthouse ticket](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2121-jpabaseequals-invokes-_key-before-checking-nullref-equality), but essentially I have an inner loop within a long-running Job which needs to test GenericModel objects for equality and in most cases the objects are equal (if they're not, then some additional work is done).  Due to the entity manager's persistence context, objects are not duplicated within the job, so references to objects that are equal are references to the same object.  That is, the object references are equal.  So the change improves performance.

The application could test for reference equality before calling equals() (which is what I effectively did to work around the problem) but that's ugly because it's the job of equals() to encapsulate equality logic, not the caller.

> Why the check should be moved?

The check should be moved because improves in the case where the object references are equal without hurting performance in the other cases.  Because of Play's and Hibernate's architecture, object references will usually be equal whenever the objects are equal.

> Why are you confident that it will not break existing applications?

I don't see how it could.  It's a simple optimization that delays computing a value until it's used and follows the pattern that was recommended in Effective Java.  It doesn't change the computed result.  I added unit tests to be sure of that.

asolntsev:
Oh, yes, you are right. I didn't catch that you actually didn't change the
behavior. Thank you for the explanation.

*@xael-fry recommended to merge this PR.*

Issue #1142, opened 2017-05-17
Make names of modules short (without version number)


asolntsev:
Currently `play deps` creates folders that contain versions numbers:

```
> play deps
> ls modules
excel-1.2.8/
fastergt-1.9.2/
guice-1.8/
liquibase-4.2/
```

It's pretty inconvenient because we need to re-configure IDE every time when we upgrade dependencies. It would be much simpler if `play deps` created modules without version numbers:
```
> play deps --shortModuleNames
> ls modules
excel/
fastergt/
guice/
liquibase/
```

asolntsev:
Done by PR https://github.com/playframework/play1/pull/1143

PR #1143, opened 2017-05-17
#1142 add parameter `--shortModuleNames` to `play deps` command

asolntsev:
it creates modules directories without version numbers

@xael-fry I personally am in doubt with parameter name:
```play deps --shortModuleNames```. 

Is it readable enough?


asolntsev:
@xael-fry Thank you! I have updated the documentation and added missing unit-tests (and squashed commits). Please check the updates.

PR #1144, opened 2017-05-22
use "application.statusKey" instead of "application.secret" …

asolntsev:
…for /@status requests

### DIFF
--- a/documentation/commands/cmd-status.txt
+++ b/documentation/commands/cmd-status.txt
@@ -8,7 +8,7 @@
 ~ 
 ~ Synopsis:
 ~ ~~~~~~~~~
-~ play status [app_path] [--url=http://...] [--secret=...]
+~ play status [app_path] [--url=http://...] [--status-key=...]
 ~
 ~ Description:
 ~ ~~~~~~~~~~~~
@@ -23,10 +23,10 @@
 ~ The script try to connect to the application at the localhost domain. If you want to monitor an application running on
 ~ a remote server, specify the application URL using this option (eg. play status --url=http://myapp.com)
 ~
-~ --secret: 
-~ The script uses the secret key to generate an authorization token. It assumes that the secret key available from the
-~ app_path/conf/application.conf is valid. If not you can provide your own secret key using this option 
-~ (eg. play status --secret=bghjT7ZG7ZGCO8)  
+~ --status-key: 
+~ The script uses the status key to generate an authorization token. It assumes that the status key available from the
+~ app_path/conf/application.conf is valid (see setting `application.statusKey`).
+~ If not you can provide your own status key using this option (eg. play status --status-key=bghjT7ZG7ZGCO8)
 ~ 
 ~ If you provide both options, you can run this command without a local application directory (app_path is not required).
 ~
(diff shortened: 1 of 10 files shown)


documentation/commands/cmd-status.txt
```diff
@@ -25,8 +25,8 @@
 ~
 ~ --secret:
```
theo: Assuming we're trying to introduce a different key for the purposes of getting status, should we change the flag from `--secret` to `--status-key` to avoid confusion?

asolntsev:
Hm... Yes, probably. Good point.

asolntsev:
Thank you @theo ! I renamed `--status` to `--status-key`. 
Additionally, I extracted "/@status" and "/@kill" part of `CorePlugin` to a dedicated class `PlayStatusPlugin` (single responsibility principle). And renamed `CorePlugin` to `EnhancerPlugin` because now it's the only job of this plugin. 

@xael-fry Please also review the changes.

asolntsev:
@xael-fry @flybyray Please review this pull request. I would like to merge it.

xael-fry:
Merged thanks

sbeigel:
How ist this supposed to work? Why did you remove the signing part in the plugin (-> server side)? I thought your goal was to not re-use the app secret but a distinct key. But I think we need to keep the signing to check the message "@status" signed w/ the "new" statusKey?!

xael-fry:
@asolntsev I let you answer this question

asolntsev:
@sbeigel @xael-fry Sorry for the long replay.
We removed the signing part because it was a potential security problem. It allowed hacker to bruteforce secretKey/statusKey.

sbeigel:
But the play status command signs the string "@status" with the application.statusKey property. Then, on the server-side, this signed message is compared to the application.statusKey property. How should this work? This will never be equal :)
To make this work right now, you have to sign the message "@status" by yourself, set this signed message as statusKey in the app.conf and pass *your* signing key as an explicit argument to the play status command... The default case in play status is useless (as it doesn't work).

asolntsev:
Sorry, I don't know what means "the default case".

We use the following command to get application status:

curl --header "authorization: YOUR-SECRET-WORD" http://localhost:9000/@status


Andrei Solntsev

2017-11-28 11:13 GMT+02:00 Sebastian Beigel <notifications@github.com>:

sbeigel:
Yes, that's what I thought :)

But this change effectively renders the play status command useless, as this command (look at the python code) signs the string "@status", either with the given secret or (that is the default) with the app.conf's property statusKey.

I understand that you want a "classic" API/secret key style. But then the play status command has to be converted to this approach as well, i.e. remove the singing part and take the passed argument *or* the app.conf's statusKey as auth message.

asolntsev:
Ok, you mean "play status" command?
I never used it.

Yes, sure: if it became broken, it should be fixed.

Anyway, I hope you also agree that it's better to break some insecure
feature than to keep it working.  :)

Andrei Solntsev

2017-11-28 11:25 GMT+02:00 Sebastian Beigel <notifications@github.com>:

sbeigel:
Yes, I do :)

I can fix the status command (in the way just described) and make a PR if you want...

asolntsev:
Sure, it would be great.
Thank you!

Andrei Solntsev

2017-11-28 11:38 GMT+02:00 Sebastian Beigel <notifications@github.com>:

sbeigel:
see #1209

PR #1149, opened 2017-05-26
[#2124] ImportError: No module named unittest

flybyray:
add unittest python module again

removed in #1134 	8ca105b


xael-fry:
PR should only contains one fix.
Will be handle in #1178
Thanks

PR #1114, opened 2017-02-23
Upgrade hibernate

asolntsev:
See issue https://github.com/playframework/play1/issues/1113

This pull request is not finally ready yet. I am sharing it just for your early reviews (if anybody is interested). I will continue working on it.

PR #1152, opened 2017-05-26
finds classes in precompiled folder too

cbxp:
currently Play 1.5 cannot find any Entity classes in precompiled mode

### DIFF
--- a/framework/src/play/classloading/ApplicationClassloader.java
+++ b/framework/src/play/classloading/ApplicationClassloader.java
@@ -262,7 +262,13 @@ public class ApplicationClassloader extends ClassLoader {
                     return res.getRealFile().toURI().toURL();
                 }
             }
-            if (Play.configuration.getProperty("play.bytecodeCache", "true").equals("true")) {
+            if (Play.usePrecompiled) {
+                File file = Play.getFile("precompiled/java/" + name);
+                if (file.exists()) {
+                    return file.toURI().toURL();
+                }
+            }
+            else if ("true".equals(Play.configuration.getProperty("play.bytecodeCache", "true"))) {
                 File f = new File(Play.tmpDir, "classes/" + name);
                 if (f.exists()) {
                     return f.toURI().toURL();


flybyray:
wait a minute, that patch makes no sense. in precompiled mode all classes are loaded before any plugin like the JPAPlugin is doing real work. if your entities are not there there is some other problem.
that fix may resolve somewhat the symptom but not the real problem. but just to late for today, i try to investigate that another time. at least it gave a hint, because those changes in applicationclassloader for the hibernate5 update is suspicious.

asolntsev:
Yes, all class are loaded. But look what method has been changed: it's
"getResource()", not "getClass()".

Unfortunately Hibernate5 call method
ClassLoader.getResource("com/my/SomeEntity.class") just to detect if entity
class exists.

flybyray:
just glad to have seen #1152
somehow it is good to describe errors in detail (stacktrace) on lighthouse or an github issue first.
then someone can make some reliable decission how to fix the real problem.
reviewing #1141 was to fast, an classloader change just for upgrading a library is suspicious.
i think the real fix could be something like #1155

asolntsev:
I like your change. It's really simpler and better solution for loading
entity classes.

Thought, my classloader change is not bad. It's a common practice that
method ClassLoader.getResource() can return *.class files. Most
classloaders do it.

flybyray:
yes might be a good change in some (odd) cases. it was not needed before, maybe it would be interessting to have a trace logging in those conditional branches to see if it would effect any test case in play. maybe it should stay a feature which is turned off by default.
but in general separation of concerns should be applied to pull requests too. a library upgrade should result in minimum changes to existing source.

ps: those many pull request from my side lastly is a result of this https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2067-fix-os-and-language-related-stuff#ticket-2067-5

flybyray:
@asolntsev 
>It's really simpler and better solution for loading entity classes.
maybe not. as i stated in https://github.com/playframework/play1/pull/1114#pullrequestreview-40684243
there is some commented todo in hibernate. the [loaded_classes](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/jpa/AvailableSettings.java#L334) is for internal use . but it is for now the only way to get the behavior as it was with 4.x hibernate

PR #1153, opened 2017-05-26
[#2127]remove regex match for remote address resolution

flybyray:
https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2127-remove-complicated-regex-based-remoteaddress-resolution


asolntsev:
@flybyray If I am not mistaken, expression `InetAddress.getByAddress(address).getHostAddress()` can be very slow (depending on network configuration). Because it tries to resolve host name and accesses DNS server. Am I right?

flybyray:
I'm afraid you're wrong.
"The InetAddress class has a cache to store successful as well as unsuccessful host name resolutions."
https://docs.oracle.com/javase/7/docs/api/

"This method doesn't block, i.e. no reverse name service lookup is performed."
https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getByAddress(byte[])

"the raw IP address in a string format."
https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getHostAddress()

this is carefully choosen and should have already the code for a long time, all those regex here was totally wrong from the beginning.

if you really care about a hostname lookup you should really call something like getCanonicalHostName
https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getCanonicalHostName()

flybyray:
but i think about the truncated ip6 scope "%" in old code.
the new code also truncates it because of raw addr .
But maybe a scope_iface or scope_id should be left as is with ip6

flybyray:
@asolntsev  from my understanding removing an existing scope is not ideal for someone who has two network interface controllers https://superuser.com/a/99753/587170
maybe with some special setup requests will not answered to local-link connections.
but i think it can be ommited because we handle (PlayHandler) only tcp and a connection is already held open, which follows its own logic to return the response.

asolntsev:
@flybyray Can you resolve conflicts with master branch?

Issue #1154, opened 2017-05-26
Play should run its own tests in precompiled mode


asolntsev:
Currently play own tests are executed in `DEV` mode. 
Thus they will not catch any errors that happen only in `PROD` mode (for example, see issue https://github.com/playframework/play1/pull/1152). 

I believe we should run play tests in `PROD` mode. It will give stronger confidence that application works in production as in test.

flybyray:
and why not just run

```
play precompile <appdir> --%test
play [auto-]test <appdir> -Dprecompiled=true
```
?

asolntsev:
@flybyray Yes, you can run tests of some application this way.
But I meant tests of Play itself. To build Play framework, we use command `play test`. And this command should run tests in `precompiled` mode - this is my point.

flybyray:
do you mean `ant test` ?

asolntsev:
Oh, sorry. Yes! I meant `ant test` (and `ant autotest`).

flybyray:
just updated the first comment. but maybe you see reason for  DEV-mode and empty db. that applies to most of the tests to. you should add jpa ddl update for prod in this case too

flybyray:
some tests have bootstrap jobs which load data, in case the data yaml cotains templates you will need to precompile them too (this was just fixed some tickets ago). but you need to specify those templates in application.conf

flybyray:
ok just checked none of the tests seem to have initial data with templates. from my point of view this is a missing code coverage

flybyray:
`@kill` will not work. you may also want to set system property "-Dwritepid=true" to have a pid file to kill the server afterwards

Fraserhardy:
Hi, We're trying to get coverage working on our play tests with JaCoCo. Are you aware of any way to get this working with Play framework? We're running 1.4.2 at the moment. I'm also interested to know if we can use ant directly to run tests too.

asolntsev:
@Fraserhardy As much as I know, it's not possible. The problem is that JaCoCo needs to modify bytecode of your application, but Play also modifies the bytecode of your application. Those two frameworks conflict with each other.

Fraserhardy:
@asolntsev Are you aware of any coverage tools that work with Play 1?

asolntsev:
The problem I mentioned is actual for any other coverage tool. So I think
it's basically not possible with Play 1.

flybyray:
Should the codecoverage only be used interactively or should it be evaluated automatically?
[...]

PR #1155, opened 2017-05-29
fix loaded entity classes in hibernate from #1114

flybyray:
revert ApplicationClassloader changes from #1114
fixing the real entity class loading problem

### DIFF
--- a/framework/src/play/classloading/ApplicationClassloader.java
+++ b/framework/src/play/classloading/ApplicationClassloader.java
@@ -255,27 +255,15 @@ public class ApplicationClassloader extends ClassLoader {
 
     @Override
     public URL getResource(String name) {
-        try {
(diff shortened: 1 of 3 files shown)


flybyray:
maybe a python test in `samples-and-tests/i-am-a-developer` could proof that the case of
#1154 #1155 is checked.
should be very easy in that sample project.
If this test would be a good idea, I would add it to this PR

asolntsev:
@flybyray I like the idea. Let's revert my classloader changes.
Yes, it would be great if you improve a python test in `samples-and-tests/i-am-a-developer`, but it's not urgent. It can be done in a separate PR.

@xael-fry I suggest to merge this PR. What do you think?

asolntsev:
@flybyray @xael-fry Thank you! Merged to master.

PR #1157, opened 2017-05-29
[#1764]feature: add advanced urlcheck

flybyray:
replacement for #706

### DIFF
--- a/framework/src/play/data/validation/URLCheck.java
+++ b/framework/src/play/data/validation/URLCheck.java
@@ -1,20 +1,66 @@
 package play.data.validation;
 
-import java.util.regex.Pattern;
 import net.sf.oval.Validator;
 import net.sf.oval.configuration.annotation.AbstractAnnotationCheck;
 import net.sf.oval.context.OValContext;
+
+import java.util.regex.Pattern;
 
 @SuppressWarnings("serial")
 public class URLCheck extends AbstractAnnotationCheck<URL> {
 
     static final String mes = "validation.url";
-    static Pattern urlPattern = Pattern.compile("^(http|https|ftp)\\://[a-zA-Z0-9\\-\\.]+\\.[a-z" +
-            "A-Z]{2,3}(:[a-zA-Z0-9]*)?/?([a-zA-Z0-9\\-\\._\\?\\,\\'/\\\\\\+&amp;%\\$#\\=~\\!])*$");
+
+    boolean tldMandatory;
+    boolean excludeLoopback;
+
+    /**
+     * big thank you to https://gist.github.com/dperini/729294
+     * well suited url pattern for most of internet routeable ips and real domains
+     */
+    static String[] regexFragments = {
+            /* 00 */ "^",
+            // protocol identifier
+            /* 01 */ "(?:(?:https?|s?ftp|rtsp|mms)://)",
+            // user:pass authentication
+            /* 02 */ "(?:\\S+(?::\\S*)?@)?",
+            /* 03 */ "(?:",
+            // IP address exclusion
+            // private & local networks
+            /* 04 */ "(?!(?:10|127)(?:\\.\\d{1,3}){3})",
+            /* 05 */ "(?!(?:169\\.254|192\\.168)(?:\\.\\d{1,3}){2})",
+            /* 06 */ "(?!172\\.(?:1[6-9]|2\\d|3[0-1])(?:\\.\\d{1,3}){2})",
+            // IP address dotted notation octets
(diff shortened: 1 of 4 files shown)


framework/src/play/data/validation/URLCheck.java
```diff
@@ -1,28 +1,84 @@
...
+            // resource path
+            /* 17 */ "(?:[/?#]\\S*)?",
+            /* 18 */ "$"
+    };
+    static Pattern urlPattern = Pattern.compile(String.join("", regexFragments), Pattern.CASE_INSENSITIVE);
 
     @Override
     public void configure(URL url) {
         setMessage(url.message());
+        this.tldMandatory = url.tldMandatory();
```
asolntsev: Do we really need to store 2 fields? Can't we just store instance of `URL`?
Like

```
public void configure(URL url) {
      this.url = url;
}
```

framework/src/play/data/validation/URLCheck.java
```diff
@@ -1,28 +1,84 @@
...
+        this.excludeLoopback = url.excludeLoopback();
     }
 
     @Override
     public boolean isSatisfied(Object validatedObject, Object value, OValContext context, Validator validator) {
         if (value == null || value.toString().length() == 0) {
             return true;
         }
-        return urlPattern.matcher(value.toString()).matches();
+        if (!tldMandatory || !excludeLoopback) {
```
asolntsev: I suggest to change order of `if`/`else`. It would make code easier to read.

```java
if (tldMandatory && excludeLoopback) {
  return urlPattern.matcher(value.toString()).matches();
}
// ... and here come all the other magic...

File: framework/src/play/data/validation/URLCheck.java [hunk 3]
```diff
@@ -1,28 +1,84 @@
...
+            //slow for special cases
+            String[] localRegexFragments = new String[regexFragments.length];
+            System.arraycopy(regexFragments, 0, localRegexFragments, 0, regexFragments.length);
```
asolntsev: Creating every time a new array seems to be too resource-consuming. And the code is not really simple.

Can't we just hold 4 regexps in 4 constants? I mean

```java
Pattern localUrlPattern_false_false = Pattern.compile(...);
Pattern localUrlPattern_false_true = Pattern.compile(...);
Pattern localUrlPattern_true_false = Pattern.compile(...);
Pattern localUrlPattern_true_true = Pattern.compile(...);
```

framework/src/play/data/validation/URLCheck.java
flybyray: I do not know if that is a good way. Maybe ask the author of OVal why he not just copies a reference  of the annotations in all of his checks.
https://github.com/sebthom/oval/tree/1.32/src/main/java/net/sf/oval/constraint

framework/src/play/data/validation/URLCheck.java
flybyray: yes maybe nicer but that is cosmetic and for me i like the most frequent return at the end of the method.

framework/src/play/data/validation/URLCheck.java
flybyray: the if part is only the special case, it will not be called every time

see URL annotation.
  `tldMandatory default true`
  `excludeLoopback default true`

flybyray:
@asolntsev  is that version better?
https://gist.github.com/flybyray/782aca447df911f0202cef9ca88f1f34

asolntsev:
@flybyray I added a comment to the gist:

May be this could be a shorter solution:

```java
private static String[] origRegex = { ... };
private static String[][] regexFragments = {
  origRegex,
  replace(origRegex, 4, "(?!(?:10)(?:\\.\\d{1,3}){3})"),
  replace(origRegex, 13, ""),
  replace(origRegex, 13, "")
};
```

flybyray:
would need something like this. but is this nicer to read?
```java
    private static String[] origRegex = { ... };
    private static IntFunction[] intf = {
            x -> (x == 4) ? "(?!(?:10)(?:\\.\\d{1,3}){3})" : null,
            x -> (x == 13) ? "" : null
    };
    private static String[][] regexFragments = {
            origRegex,
            replace(origRegex.clone(), intf[0]),
            replace(origRegex.clone(), intf[1]),
            replace(replace(origRegex.clone(), intf[0]), intf[1])};
    private static String[] replace(String[] array, IntFunction<? extends String> generator) {
        for (int i = 0; i < array.length; i++) {
            final String object = generator.apply(i);
            if (object != null) array[i] = object;
        }
        return array;
    }
```

PR #1159, opened 2017-05-31
[#2089 ] upgrade bouncycastle

flybyray:
this is a proposed pull request needs at least some other pr merged first. not all are required but i added them to test more easily from a windows system.

it adds a new tests for ssl certificate usage


asolntsev:
@flybyray Which "other" PR do you mean?

flybyray:
#1155 

all others are not needed for this bouncycastle update. i just added them for myself testing it on windows os system. hence they are only needed if you like play be portable to windows os.

since you merged #1155 the pr #1159 could be an independent pr with a single commit

flybyray:
ok looks good!
code/plugins using for example https://www.license4j.com/ (newer bouncycastle jars) do not conflict with play runtime jars

asolntsev:
@xael-fry To be honest, I am not able to understand if this is good change or not. 
So I suggest to trust the PR author and merge it.

framework/dependencies.yml
```diff
@@ -42,7 +42,8 @@ require:    &allDependencies
     - oauth.signpost -> signpost-core 1.2.1.2
     - org.apache.geronimo.specs -> geronimo-servlet_2.5_spec 1.2
     - org.apache.ivy -> ivy 2.4.0
-    - org.bouncycastle -> bcprov-jdk15 1.46
+    - org.bouncycastle -> bcprov-jdk15on 157
```
flybyray: True. But on the original homepage you will find `bcprov-jdk15on-157.jar` without a period dot between numbers. see https://www.bouncycastle.org/latest_releases.html
I am not sure if all Play! dependencies are really from maven. That would be nice because of checksums and sometimes additional signatures.

flybyray:
changed the upgraded bouncycastle dependencies to its maven versions (adding the point into version number)

asolntsev:
@flybyray Thank you! Merged.

Issue #1133, opened 2017-05-02
Allow customization of @CacheFor cache key


tahseenarticle:
At the moment when you annotate an Controller method with CacheFor. Play framework generate the cache key as follow in ActionInvoker.java line 152:

```
cacheKey = "urlcache:" + request.url + request.querystring;
```

CacheFor annotation does not allow customisation of this cache key. This is useful in case you your controller shows different page based on the session. For example if you viewing different language version of the website the URL will be the same but the locale information is in the session. Having a way to customise this cache key helps in lot different cases of caching.

My proposal is to extend the @CacheFor as follow:

```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
public @interface CacheFor {
    String value() default "1h";
    String id() default "";
    Class<? extends CacheKeyGenerator> generator() default DefaultCacheKeyGenerator.class;
}
```
Where CacheKeyGenerator is a simple interface that generate the cache key.

```java
public interface CacheKeyGenerator {
	public String generate(Request request);
}
```
```java
public class DefaultCacheKeyGenerator implements CacheKeyGenerator {
	@Override
	public String generate(Request request) {
		return "urlcache:" + request.url + request.querystring;
	}
}
```

Then we can modify the InvokerAction.invoke to generate the cache key based on the CacheKeyGenerator.

You can use a custom CacheKeyGenerator as follow:

```
@CacheFor(value = "1h", generator=SessionBasedCacheKeyGenerator.class)
```
If you do not specify the ```generator``` the default key generator will be used as usual.

```
@CacheFor(value = "1h")
```

PS: I am working on a pull request to this issue

asolntsev:
@tahseenarticle Good luck! Let us know when the PR is ready.

asolntsev:
@tahseenarticle Do you have some updates? Could you make a pull request? Can I help with this?

asolntsev:
Implemented in PR #1166

PR #1166, opened 2017-06-19
Allow customization of @CacheFor cache key

tahseenarticle:
Please see issue #1133

At the moment when you annotate an Controller method with CacheFor. Play framework generate the cache key as follow in ActionInvoker.java line 152:

```
cacheKey = "urlcache:" + request.url + request.querystring;
```

CacheFor annotation does not allow customisation of this cache key. This is useful in case you your controller shows different page based on the session. For example if you viewing different language version of the website the URL will be the same but the locale information is in the session. Having a way to customise this cache key helps in lot different cases of caching.

My proposal is to extend the @CacheFor as follow:

```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
public @interface CacheFor {
    String value() default "1h";
    String id() default "";
    Class<? extends CacheKeyGenerator> generator() default DefaultCacheKeyGenerator.class;
}
```
Where CacheKeyGenerator is a simple interface that generate the cache key.

```java
public interface CacheKeyGenerator {
	public String generate(Request request);
}
```
```java
public class DefaultCacheKeyGenerator implements CacheKeyGenerator {
	@Override
	public String generate(Request request) {
		return "urlcache:" + request.url + request.querystring;
	}
}
```

Then we can modify the InvokerAction.invoke to generate the cache key based on the CacheKeyGenerator.

You can use a custom CacheKeyGenerator as follow:

```
@CacheFor(value = "1h", generator=SessionBasedCacheKeyGenerator.class)
```
If you do not specify the ```generator``` the default key generator will be used as usual.

```
@CacheFor(value = "1h")
```


framework/src/play/mvc/ActionInvoker.java
```diff
@@ -147,11 +147,15 @@ public static void invoke(Http.Request request, Http.Response response) {
 
                 // Check the cache (only for GET or HEAD)
                 if ((request.method.equals("GET") || request.method.equals("HEAD")) && actionMethod.isAnnotationPresent(CacheFor.class)) {
-                    cacheKey = actionMethod.getAnnotation(CacheFor.class).id();
+                    CacheFor cacheFor = actionMethod.getAnnotation(CacheFor.class);;
+                    cacheKey = cacheFor.id();
                     if ("".equals(cacheKey)) {
-                        cacheKey = "urlcache:" + request.url + request.querystring;
+                        // Generate a cache key for this request
+                        cacheKey = cacheFor.generator().newInstance().generate(request);
```
asolntsev: Can't it cause a bigger memory footprint because of creating a new instance of generator every time?

asolntsev:
@xael-fry Please review the PR. From a code quality, it seems ok. Though, probably a little bit overkill... Or not?

framework/src/play/mvc/ActionInvoker.java
tahseenarticle: There two options here:
1) We can maintain a cache of CacheKeyGenerator instances in the ActionInvoker.
2) Push the creation of generator to be application developer responsibility by added a get() method to the CacheKeyGenerator interface.

public interface CacheKeyGenerator {
	public String generate(Request request);
        public CacheKeyGenerator get();
}

Line 154 will can rewitten as:

cacheKey = cacheFor.generator().get().generate(request);

tahseenarticle: DefaultCacheKeyGenerator can implement a singleton 

```Java

public class DefaultCacheKeyGenerator implements CacheKeyGenerator {
        private final static CacheKeyGenerator instance;
	@Override
	public String generate(Request request) {
		return "urlcache:" + request.url + request.querystring;
	}

       public CacheKeyGenerator get() {
               if(instance == null) {
                        instance = new DefaultCacheKeyGenerator();
               }
               return  instance;
       }
}
```

tahseenarticle: I can make quick these changes if this is what is required to get this pull request out of the door?
I am also all ears for any other solutions.

asolntsev: @tahseenarticle I am not really sure if we need to change anything. I just asked :)
Probably it would be a premature optimization. 
What I actually am waiting is @xael-fry review.

PR #1169, opened 2017-07-03
[#2130] remove annoying log messages from autotest

flybyray:


xael-fry:
Merged in master
thanks

PR #1075, opened 2016-12-29
ticket 2073: add simple test for method IO.copyDirectory()

asolntsev:
just a simple test showing that IO.copyDirectory() works as expected

(see ticket https://play.lighthouseapp.com/projects/57987/tickets/2073)

PR #1171, opened 2017-07-03
ticket 2073: add simple test for method IO.copyDirectory()

flybyray:
#1075 testing

ok it works only with java.nio.Files https://stackoverflow.com/a/21987929

The Problem in the test is the `Ü` Character in the filename.
On Windows for example the utf-8 string will translate to `Ãœ` when accessing the file system.
```
INFORMATION: cp1252: C:\dev\src\github.com\playframework\play1\framework\test-classes\play\libs\Anhang Performance-Bericht_ Ãœbersicht Â§18 Bundesbank gesetz Meldepflichten.PDF
Jul 04, 2017 5:03:31 AM play.Logger info
INFORMATION: cp1252: C:\dev\src\github.com\playframework\play1\framework\test-classes\play\libs\Anhang Performance-Bericht_ UÌˆbersicht Â§18 Bundesbank gesetz Meldepflichten.PDF
```

But if you list the files with java and print it out it will be `UÌˆ`.
```
        Path path1 = target.listFiles()[1].toPath().toAbsolutePath();
        String anotherString2 = path1.toString();
        cp1252 = new String(anotherString2.getBytes("UTF-8"), "WINDOWS-1252");

```


flybyray:
The best way would be to fix legacy code in the play framework https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2132-replace-legacy-file-io-code

PR #1178, opened 2017-08-23
[#2124] fix(python): add missing unitest lib

xael-fry:
[lighthouse #2124](https://play.lighthouseapp.com/projects/57987/tickets/2124)


xael-fry:
It is not a 3rd-party  lib, it is an optional module of python 2.7 present in installer.
So I don see any problem here

asolntsev:
@xael-fry I still don't understand :)
But if you don't see problems, we can merge it.

xael-fry:
The lib was present before, but I accidentally remove it wen updating python.
As it is a Python package lib, we just need to keep update python

PR #1180, opened 2017-08-23
Make it possible to change session store implementation

cbxp:
There are some problems with the default cookie-based sessions, including:
* It is very easy to overwrite changed session in concurrent AJAX request with unpredictable result
* Session replay attacks are possible by sending previous version of the session until it expires

This PR makes it possible to implement other backends for session storage, e.g. cache or database.
This is similar to CacheImpl, where new caching implementaions can be created.

The default implementation is CookieSessionStore that extracted as-is from Scope.Session class. All Scope.Session API is left 100% unchanged, so no breaking changes.

No new implementations are provided as of now, but we can create a separate PR with cache-based implementation.

### DIFF
--- a/framework/src/play/mvc/Scope.java
+++ b/framework/src/play/mvc/Scope.java
@@ -1,13 +1,5 @@
 package play.mvc;
 
-import java.lang.annotation.Annotation;
-import java.net.URLEncoder;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.UUID;
-
 import play.Logger;
 import play.Play;
 import play.data.binding.Binder;
@@ -20,8 +12,11 @@ import play.exceptions.UnexpectedException;
 import play.i18n.Messages;
 import play.libs.Codec;
 import play.libs.Crypto;
-import play.libs.Time;
 import play.utils.Utils;
+
+import java.lang.annotation.Annotation;
+import java.net.URLEncoder;
+import java.util.*;
 
 /**
  * All application Scopes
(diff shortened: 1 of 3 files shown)


asolntsev:
@xael-fry I think the change is good. Recommended to merge.

framework/src/play/mvc/Scope.java
```diff
@@ -37,6 +34,19 @@
     public static final boolean SESSION_SEND_ONLY_IF_CHANGED = Play.configuration
             .getProperty("application.session.sendOnlyIfChanged", "false").toLowerCase().equals("true");
 
+    public static SessionStore sessionStore = createSessionStore();
+
+    private static SessionStore createSessionStore() {
+        String sessionStoreParameter = Play.configuration.getProperty("application.session.store", "cookie");
```
theo: Would be better if we just have application.session.storeClass which takes in the class instead of relying on some naming convention that has a "SessionStore" suffix.

asolntsev: @theo @angryziber @cbxp I agree.

angryziber: I thought the simpler parameter is more consistent with other parameters, e.g. memcached

angryziber: I thought the simpler parameter is more consistent with other parameters, e.g. memcached

cbxp:
I have made the requested change. How about a merge?

cbxp:
Note: build actually passed. For some reason it status was not updated here

xael-fry:
@asolntsev Maybe documentation of  `application.session.storeClass` should have been great before merging it

asolntsev:
@xael-fry Good catch. I will add it to the documentation.

xael-fry:
Should the name be `application.session.store.class` ?

cbxp:
It seems unlikely that there will be more general store parameters.
If any additional store would have its own parameters, they better be in its own namespace.
So we suggest to leave it as `application.session.storeClass`

PR #1181, opened 2017-08-25
Avoid restoring of flash and session multiple times per request

cbxp:
This fixes a regression introduced somewhere in Play 1.3

Session and flash will be restored only once per request (it is at least 3 times currently)
Depends on PR #1180, should be merged after it


asolntsev:
@xael-fry It seems to be a good PR. Recommended to merge.

framework/src/play/mvc/Router.java
```diff
@@ -354,13 +349,9 @@ public static void routeOnlyStatic(Http.Request request) {
...
-            } catch (Throwable t) {
-                if (t instanceof RenderStatic) {
-                    throw (RenderStatic) t;
-                }
-                if (t instanceof NotFound) {
-                    throw (NotFound) t;
-                }
+            } catch (RenderStatic | NotFound e) {
+                throw e;
+            } catch (Throwable ignore) {
```
xael-fry: Why did you put a silent catch here, I don't think it is a good idea

asolntsev: @xael-fry This is exactly how the previous code worked. 
@cbxp But I agree, it would be better to add logging for those exceptions.

xael-fry: @asolntsev You are right sorry, maybe check that adding a log will not generate too many logs, it is maybe a normal behavior.

cbxp: @xael-fry As this PR doesn't change the behaviour here, I suggest not to introduce more changes, but make them separately if needed. This change was suggested by the IDE and just makes it more obvious that other Throwables are just ignored.

asolntsev: @xael-fry So, can we merge this PR?

xael-fry:
Merged and pushed in master
Thanks @cbxp  @asolntsev

JohannesBeranek:
Looking at changed files this doesn't look like it does anything other than adding an unused import? Maybe you forgot to merge something or something dropped along the line? Or does github maybe just not show it?

asolntsev:
Yes, the diff looks strange. 
Probably here lies the explanation:
> Depends on PR #1180, should be merged after it

PR #1186, opened 2017-09-04
[#2138]enhance the function for the injector (Add the injectable interface)

zhaoyin:
Hi,the Injector inject beans in controllers,now I enhance the function of the injector

[#Lighthouse 2138](https://play.lighthouseapp.com/projects/57987-play-framework/tickets/2138)

### DIFF
--- a/framework/src/play/inject/Injectable.java
+++ b/framework/src/play/inject/Injectable.java
@@ -1,0 +1,18 @@
+package play.inject;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.List;
+import javax.inject.Inject;
+import play.Play;
+import play.classloading.enhancers.ControllersEnhancer.ControllerSupport;
+import play.jobs.Job;
+import play.mvc.Mailer;
+
+/**
+ * if you want to inject beans anywhere,please implement the interface.
+ *
+ */
+public interface Injectable {
+}
--- a/framework/src/play/inject/Injectable.java
+++ b/framework/src/play/inject/Injectable.java
@@ -1,15 +1,5 @@
 package play.inject;
 
-import java.lang.reflect.Field;
-import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.List;
-import javax.inject.Inject;
-import play.Play;
-import play.classloading.enhancers.ControllersEnhancer.ControllerSupport;
-import play.jobs.Job;
-import play.mvc.Mailer;
-
 /**
  * if you want to inject beans anywhere,please implement the interface.
  *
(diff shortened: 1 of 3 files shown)


asolntsev:
@zhaoyin Thank you!
This is basically a good change, except two nuances.

1. I guess "Injectable" is a wrong term. "Injectable" means the thing **being injected**. 
What you need is "InjectionClient" or probably "RequiresInjection" (see https://en.wikipedia.org/wiki/Dependency_injection).
2. I suggest to create an annotation instead of interface. It seems to be a better design:

```
@RequiresInjection
public class MyJob extends Job {
  @Inject MyService service;
}
```

zhaoyin:
hi,@asolntsev,for example,i have a personDao,a personService,now ,playframework supports inject a personService in controller,but it can't support inject a dao in service,so i add a interface to support this case

zhaoyin:
hi,@asolntsev,I know that,but in playframework,the Injector class in inject package,the original method `` public static void inject(BeanSource source)`` inject beans in controllers  that implements ControllerSupport,ControllerSupport is a interface ,i use the feature ,so i use the injectable interface instead of an annotation to the minimal changes.i think it's ok,and you ?

zhaoyin:
and when ever was  the milestone 1.5 release publishing ？

asolntsev:
@zhaoyin I agree that "minimal changes" is a valid concern.
But in this case, using annotation is also very simple:

```
classes.addAll(Play.classloader.getAnnotatedClasses(RequiresInjection.class));
```

Byt the way, why do you want to inject dependencies somewhere but controllers and jobs?
Typically it's not needed. The main entry point is controller. Controller injects its dependencies, and every of its dependencies injects its own dependencies, and so on. No need for any custom injection.

zhaoyin:
@asolntsev ,many projects have Layered development，every layer should be a interface for it's upper layer but only for controller，especially the project got bigger than before.the some layer of the project may be a part of other project.so i think the injector shouldn't for the controller...And you?

asolntsev:
@xael-fry I recommend to merge this PR.

asolntsev:
@zhaoyin Please also review my next pull request: https://github.com/playframework/play1/pull/1198

Probably it will also be useful for you?

asolntsev:
@zhaoyin Please answer: how exactly do you use `play.inject.Inject` class?
As I see from Play source code, this class is never used.

zhaoyin:
defined a plugin for Inject(e.x. Spring,Guice),in the plugin ,I use the method ``public static void inject(BeanSource source)`` and the annotation ``play.inject.Inject``

asolntsev:
@zhaoyin Thank you for the answer. We also use `play-guice` plugin. 

Then I don't understand why you (and me) need the class `play.inject.Inject`. Why can't you just do the same thing in your plugin?

PR #1192, opened 2017-09-28
Update Configuration.java

gaoquanbing-ausxin:
when you set “db.factory=....“ in application.conf will lead error


sergeyrolich:
+1

asolntsev:
@gaoquanbing-ausxin Thank you for the fix!

PR #1205, opened 2017-11-02
Add db.factory to the dbProperties list

AlexLaviolette:
If you try to add the db.factory configuration option, currently the DBPlugin thinks this is a named database and not just a conf option. It was missing from the dbProperties list in Configuration.java

### DIFF
--- a/framework/src/play/db/Configuration.java
+++ b/framework/src/play/db/Configuration.java
@@ -38,7 +38,7 @@ public class Configuration {
         Pattern pattern = new jregex.Pattern(DB_CONFIG_PATTERN);
 
         // List of properties with 2 words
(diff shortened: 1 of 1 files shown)


framework/src/play/db/Configuration.java
```diff
@@ -38,7 +38,7 @@ public Configuration(String configurationName) {
         Pattern pattern = new jregex.Pattern(DB_CONFIG_PATTERN);
 
         // List of properties with 2 words
-        List<String> dbProperties = Arrays.asList("db.driver", "db.url", "db.user", "db.pass", "db.isolation", "db.destroyMethod",
+        List<String> dbProperties = Arrays.asList("db.driver", "db.url", "db.user", "db.pass", "db.factory", "db.isolation", "db.destroyMethod",
```
xael-fry: `db.factory` is already in the list, check next line

PR #1208, opened 2017-11-22
[#2140] Make FunctionalTest compatible with Response.writeChunk()

davidcostanzo:
This patch updates FunctionalTest to register an onWriteChunk action on the Response objects which it creates so that controller actions which call Response.writeChunk(Object) do not fail with an unhandled exception when invoked from GET().

This patch also includes a regression test, added to FunctionalTestTest, which calls an existing controller action in Application.


asolntsev:
@xael-fry The PR seems to be ok. Recommended to merge.

xael-fry:
Merged in master thanks

Issue #1212, opened 2017-12-25
Wrong request-body decoding for content-type "application/x-www-form-urlencoded"


AlinaDidenko:
Hi, 
there is request with content-type "application/x-www-form-urlencoded; charset=utf-8" and cyrillic symbols in body. After update to play-1.5.0 from play-1.2.7.2 controller params contain invalid character sequences instead cyrillic words. This is due to an error in org.apache.common.codec.net.URLCodec#decode(java.lang.String, java.lang.String) (used in play/data/parsing/UrlEncodedParser.java:132). 
For example:
```
final String str = "Привет";
System.out.println("java.net.URLDecoder: " + URLDecoder.decode(str, "UTF-8")); // in play 1.2.7.2
System.out.println("org.apache.commons.codec.net.URLCodec: " + new URLCodec().decode(str, "UTF-8")); // in play 1.5.0
```
Output:
```
java.net.URLDecoder: Привет
org.apache.commons.codec.net.URLCodec: ??????
```
This error lies in the incorrect receipt of bytes: string.getBytes(Charsets.US_ASCII); 
So this decision works only for standard ascii-symbols. 
Please fix the error.

asolntsev:
@AlinaDidenko Can you prepare a pull request?

AlinaDidenko:
@asolntsev Yes, we will try to fix it

GordeevArt:
Why no review? Why not in the 1.5.1 version?

asolntsev:
@GordeevArt is there a pull request that we could review?

GordeevArt:
@asolntsev https://github.com/playframework/play1/pull/1213

xael-fry:
Merge #1213 
Thanks

PR #1213, opened 2018-01-11
[#1212] Fix non-Latin characters decode in UrlEncodedParser

AlinaDidenko:
This PR fixes the decoding of a query string from requests with content-type "application/x-www-form-urlencoded". Error detected in org.apache.common.codec.net.URLCodec so it is replaced by java.net.URLDecoder.


xael-fry:
@asolntsev thanks

Issue #1221, opened 2018-02-26
[#2155] Silent or suppressed errors in unittest PluginCollectionTest#skipsDuplicatePlugins


flybyray:
02a75fb in PR #1138
Introduced resource "custom-play.plugins.duplicate" which depends on CorePlugin

15874d0 in PR #1144
Splitted CorePlugin into two Plugins with new Names and Location without modifying custom-play.plugins.duplicate.

Ant task unittest should fail because resource "custom-play.plugins.duplicate" file will not be copied ( see https://github.com/playframework/play1/blob/master/framework/build.xml#L397 ).
Hence the PR #1138 does not really test what it should test.

xael-fry:
Merged and pushed in master
Thanks

PR #1222, opened 2018-02-26
[#2155] Silent or suppressed errors in unittest PluginCollectionTest#skipsDuplicatePlugins

flybyray:
see #1221


flybyray:
the renaming of `custom-play.plugins.duplicate` is because of

https://github.com/playframework/play1/blob/master/framework/build.xml#L397 

will not copy *.duplicate files.

PR #1226, opened 2018-03-24
#1237 Use secure websocket protocol for secure pages

sbeigel:
Switch from HTTPS to WSS (like we already do it for HTTP to WS) to support secure websocket connections. At least Chrome doesn't allow "mixed content" like using WS for HTTPS pages.


xael-fry:
Cf issue #1237

xael-fry:
Merged and pushed in master.
Thanks

PR #1231, opened 2018-04-30
[#2171] Action context access from error templates for routes-defined 404 actions

sant0s:
This fix removes action context resetting making action context items available to error templates associated with routes-defined 404 actions.


sant0s:
@xael-fry For some reason, ticket 2170/PR 1230 went through in Play! 1.5.2, but related ticket 2171/PR 1231 (this one) didn't. I've just rebased this pull request's branch onto the latest master and re-tested successfully i.e. checked that the issues reported in the ticket show up/do not show up before/after this fix is applied. Could this pull request be included in the next Play! release?

PR #1234, opened 2018-04-30
[#2172] Apply NotFound result thrown during action resolution failure to response

sant0s:


sant0s:
Although this change makes unavailable dynamic resources look (correctly) unavailable to functional tests (HTTP 404 is returned instead of HTTP 200), it breaks serving of valid/existing staticDir/staticFile ones (HTTP 404 is returned instead of HTTP 200).
Hopefully, a Play maintainer will be able to fix (Lighthouse) ticket 2172 so that correct REST-compliant HTTP responses are returned to functional tests in the same way they are returned to non-testing scenarios.

PR #1242, opened 2018-06-21
Add support for navigating XML with namespaces.

ludovicofischer:
Add method overloads that let you generate documents that you can naviagate with XPath while taking XML namespaces into account.
This adds method to get namespace-aware XML documents from strings, files, input streams and request bodies.
Without this, navigating the documents with XPath always returns null if namespaces are taken into account. I’ve added a few tests to demonstrate the issue.


xael-fry:
Merged in master
Thanks @ludovicofischer

Issue #1200, opened 2017-10-22
Does play 1 support java 9?


alekseysotnikov:

asolntsev:
Hi Aleksei!
I didn't even try yet. I cannot even run unit-tests in my current project
with Java9, so I am skeptical about that.

Did you try to run Play application with Java 9?

Andrei Solntsev

2017-10-22 21:27 GMT+03:00 Aleksei Sotnikov <notifications@github.com>:

AlexLaviolette:
It appears that Play 1 does not currently support Java 9. At the very least, ApplicationCompiler.java will need to be updated to support the new version.

tazmaniax:
I'm getting the following error with continuations on Java 9 which might be related to the above point about ApplicationCompiler and/or JavaFlow...
```
10:26:32,015 [play-thread-1] ERROR ~ 

@75ph95fhb
Internal Server Error (500) for request GET /oauth/callback?state=2&scope=read_write&code=ac_BiBi56NS6A7gcQ7EtYgaCG1SUksAl0AB

await/Continuations error
A await/Continuations error occurred : Cannot use await/continuations when not all application classes on the callstack are properly enhanced. The following class is not enhanced: jdk.internal.reflect.NativeMethodAccessorImpl

play.exceptions.ContinuationsException: Cannot use await/continuations when not all application classes on the callstack are properly enhanced. The following class is not enhanced: jdk.internal.reflect.NativeMethodAccessorImpl
	at play.mvc.Controller.verifyContinuationsEnhancement(Controller.java:1244)
	at play.mvc.Controller.await(Controller.java:1206)
	at controllers.StripeListener.oAuthCallback(StripeListener.java:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at play.mvc.ActionInvoker.invokeWithContinuation(ActionInvoker.java:538)
	at play.mvc.ActionInvoker.invoke(ActionInvoker.java:475)
	at play.mvc.ActionInvoker.invokeControllerMethod(ActionInvoker.java:469)
	at play.mvc.ActionInvoker.invokeControllerMethod(ActionInvoker.java:436)
[...]

PR #1245, opened 2018-07-17
fix(tests): issue #1200, added testing compatibility for jdk9+

SomelifeDev:
This solves the "await/Continuations error" when trying to run tests in if you're using jdk9+

### DIFF
--- a/framework/src/play/mvc/Controller.java
+++ b/framework/src/play/mvc/Controller.java
@@ -1234,7 +1234,7 @@ public class Controller implements PlayController, ControllerSupport, LocalVaria
                 }
 
                 if (haveSeenFirstApplicationClass) {
-                    if (className.startsWith("sun.") || className.startsWith("play.")) {
+                    if (shouldBeCheckedForEnhancement(className)) {
                         // we're back into the play framework code...
                         return; // done checking
                     } else {
@@ -1250,6 +1250,17 @@ public class Controller implements PlayController, ControllerSupport, LocalVaria
             }
 
         }
+    }
+
+
+    /**
+     * Checks if the classname is from the jdk, sun or play package and therefore should not be checked for enhancement
+     * @param String className
+     * @return boolean
+     */
+    static boolean shouldBeCheckedForEnhancement(String className)
+    {
+        return className.startsWith("jdk.") || className.startsWith("sun.") || className.startsWith("play.");
     }
 
     protected static <T> void await(Future<T> future, F.Action<T> callback) {
--- a/framework/test-src/play/mvc/ControllerTest.java
+++ b/framework/test-src/play/mvc/ControllerTest.java
@@ -1,0 +1,14 @@
+package play.mvc;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+import static play.mvc.Controller.shouldBeCheckedForEnhancement;
+
+public class ControllerTest{
+    @Test
+    public void jdkAndPlayClassesShouldNeverBeenCheckedForEnhancement() {
+        assertTrue(shouldBeCheckedForEnhancement("sun.blah"));
+        assertTrue(shouldBeCheckedForEnhancement("play.foo"));
+        assertFalse(shouldBeCheckedForEnhancement("com.bar"));
+    }
+}


tomparle:
Hi SomelifeDev,

I confirm this solves the `Cannot use await/continuations when not all application classes on the callstack are properly enhanced. The following class is not enhanced: jdk.internal.reflect.NativeMethodAccessorImpl` error when using jdk9.
I also confirmed that it worked with jdk10.

There is an additional warning that I'm not qualified enough to say if it's important or not : 
`WARNING: Illegal reflective access by javassist.util.proxy.SecurityActions (file:.../play-1.x/framework/lib/javassist-3.21.0-GA.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)`

So, thanks a lot ! I am looking forward this fix to be merged since it would finally allows us to use Play 1.x with more recent versions of JDK.

tazmaniax:
I would say that I'm equally unqualified to resolve :) but looking at the javassist release notes, [v3.22](https://github.com/jboss-javassist/javassist/blob/master/Readme.html#L297) introduced support for JDK 9 and the latest GA release is [v3.23.1](https://github.com/jboss-javassist/javassist/releases/tag/rel_3_23_1_ga). If it's possible to upgrade javassist then that's probably worth doing. If that's not possible or doesn't resolve it then it's possible to open up the required modules for javaasist which can be found in this [migration guide](https://static.rainfocus.com/oracle/oow17/sess/1491919829459001saEk/PF/migrating%20to%20java%209%20modules_1507330319706001mF5M.pdf) - see slide 16 and 31.

tomparle:
Hi Chris,

thank you for the quick reply !
Indeed, upgrading to the latest version of javassist did the trick !
I created the corresponding pull request which only required a small change : https://github.com/playframework/play1/pull/1247

Please tell me if the pull request is ok or if I should have done something differently, I'm not that much used to contribute, yet !
Tom

tazmaniax:
Connecting this to #1200

tomparle:
@SomelifeDev it's not that we did not bother, but personally I had no idea of how to fix this initially ! So thank you for having unblocked the situation.
Regarding to the corresponding test case, while it seems that there is no test class for Controller specifically, among all tests in error, there is the test `ContinuationTest` that explicitely failed with jdk9+, but passed with SomelifeDev's fix.
So maybe I guess we may not need to add another test specifically, what do you think ?

asolntsev:
@tomparle @SomelifeDev  Regarding a unit-test. I mean, in normal situation I would
1. extract `className.startsWith("sun.") || className.startsWith("play.")` to a separate method in `Controller` class. Say, 
```java
static boolean shouldNotBeEnhanced(String className) {
  return className.startsWith("sun.") || className.startsWith("play.");
}
```
2. I would create a unit-test `ControllerTest.java` with couple of asserts:
```java
@Test
public void jdkAndPlayClassesShouldNeverBeenEnhanced() {
  assertTrue(shouldNotBeEnhanced("sun.blah"));
  assertTrue(shouldNotBeEnhanced("play.foo"));
  assertFalse(shouldNotBeEnhanced("com.bar"));
}
```
But in this case I guess we can omit this step because there is so much code that is not covered by unit-tests at all.

tomparle:
@asolntsev  thanks for your test suggestion, sounds right to me, especially because it enlightens the "ignore for enhancement" mechanism, which is otherwise kind of hidden in Play internals.
@SomelifeDev what do you think ? it should be quick to add this test.
Otherwise I will add it next week (not sure if I can contribute to the pull request of another user ?)

SomelifeDev:
@asolntsev and @tomparle
Done (after a couple massive monday morning brain farts).
Was a good idea and even if lot of tests are missing, it's the occasion to start on the ControllerTest class anyway. Plus allowed me to finally take the time to properly set things up on my side...

asolntsev:
@SomelifeDev Great work, thank you!
I have two more wishes:
1. Method `shouldBeCheckedForEnhancement` could be simplified: 
```
static boolean shouldBeCheckedForEnhancement(String className) {
  return className.startsWith("jdk.") || className.startsWith("sun.") || className.startsWith("play.");
}
```
2. Could you *squach* your N commits into one commit? It would simplify git history tree.

SomelifeDev:
@asolntsev done and done.

asolntsev:
@SomelifeDev Thank you! Merged.

tomparle:
@SomelifeDev thank you for the fix and having taken into account @asolntsev 's suggestions !

framework/src/play/mvc/Controller.java
```diff
@@ -1252,6 +1252,17 @@ private static void verifyContinuationsEnhancement() {
         }
     }
 
+
+    /**
+     * Checks if the classname is from the jdk, sun or play package and therefore should not be checked for enhancement
```
tazmaniax: Sorry for being late to the party but this comment feels counter intuitive.  It states that if the class name package is jdk, sun or play then it should NOT be checked for enhancement. As the method returns true if any of those packages match then shouldn't the method be shouldNotBeCheckedForEnhancement() ? Or am I having a brain fart here :)

SomelifeDev:
@tazmaniax 
You're 100% correct

framework/src/play/mvc/Controller.java
asolntsev: Yes, good point. I already fixed it in master branch.

PR #1246, opened 2018-07-18
[#1849] Exceptions from controllers are not detected in FunctionalTests

Fraserhardy:
Lighthouse #1849 from PR #778 

Changes have been ported over from an existing PR that became out of date. 
This fixes a pretty critical issue where functional tests will pass even if the controller being called throws an exception. 

The new changes actually throw the exception up to the tests so that it can be handled within the test. Expected exceptions can be checked for correctly in this way.


samples-and-tests/just-test-cases/test/FunctionalTestTest.java
```diff
@@ -120,13 +121,70 @@ public void canGetRenderArgs() {
         User u = (User) renderArgs("u");
         assertEquals("Guillaume", u.name);
     }
-    
-    @Test
+
+    @Test(expected = RenderStatic.class)
```
Fraserhardy: The changes to how exceptions are thrown back to tests has resulted in RenderStatic.class being thrown back up to the tests in this case. 

This change passes the test but I'm not 100% convinced this is the correct approach. Could someone who has more experience in how the static files are handled take a look at this please.

framework/src/play/test/FunctionalTest.java
```diff
@@ -377,6 +399,21 @@ public InvocationContext getInvocationContext() {
...
 
+    private static RuntimeException unwrapOriginalException(final ExecutionException e) {
+        // Check if the original exceptions fits the usual patterns. If yes, throw the very
+        // original runtime exception
+        final Throwable executionCause = e.getCause();
+        if (executionCause != null
+              && (executionCause instanceof JavaExecutionException || executionCause instanceof UnexpectedException)) {
+            final Throwable originalCause = executionCause.getCause();
+            if (originalCause != null && originalCause instanceof RuntimeException) {
+                throw (RuntimeException) originalCause;
```
asolntsev: This is bug. Change `throw (RuntimeException) originalCause;` to `return (RuntimeException) originalCause;`

samples-and-tests/just-test-cases/test/FunctionalTestTest.java
asolntsev: @Fraserhardy @hsch Certainly, it's a bad idea to expect `RenderStatic.class` in this test. But it's not your fault.

There is two problems with this test:
1. it WAS incorrect before your change. It actually didn't test anything. If only you add a line `assertContentEquals("test", response);` to this test, it would fail.

2. Bad design of method `play.mvc.Router.Route#matches(java.lang.String, java.lang.String, java.lang.String, java.lang.String)`: it has size effect. Instead of returning `Map<String, String>`, it sometimes throws `RenderStatic`. And `PlayHandler` handles it as a special case: catches `RenderStatic` and sends file content asynchronously to the browser. Disgusting. :(

What I suggest is making a similar "special case" in `FunctionalTest`:

```java
        catch (ExecutionException e) {
          RuntimeException originalException = unwrapOriginalException(e);
          if (originalException instanceof RenderStatic) {
            response.status = 200;
            response.direct = ((RenderStatic) originalException).file;
          }
          else {
              throw originalException;
          }
        }
```

and adding a method in `FuncationalTest`:

```java
    public static void assertIsStaticFile(Response response, String filePath) {
        assertIsOk(response);

        String file = (String) response.direct;
        assertEquals(filePath, file);
    }
```

Finally, this test should be rewritten:

```java
    @Test
    public void testGettingStaticFile() {
	      Response response = GET("/public/session.test?req=1");
	      assertIsStaticFile(response, "public/session.test");
    }
```

asolntsev:
@Fraserhardy @hsch I merged your pull request and then implemented my comments in a separate PR. You can see the final version in master branch now.

Fraserhardy:
@asolntsev Thanks for your help on this. Much appreciated.

Fraserhardy:
@asolntsev Now this has been merged what steps are required to get this into a new version? I know we've just missed 1.5.1. I noticed there are a few vulnerable dependencies that have just been updated too so perhaps another release is due soon.

asolntsev:
@Fraserhardy Unfortunately, it's not an automated process. Typically @xael-fry does the release. It includes some additional work like preparing a documentation etc. 

@xael-fry Where do you plan to make a new release?

xael-fry:
@asolntsev In september, but this one should not have been merged, modifications should have been combined in one commit.
I would have to clean it up

PR #1247, opened 2018-07-20
upgraded javassist to 3.23.1-GA to suppress warning with JDK9+

tomparle:
In addition to #1245 that fix JDK9+ compatibility (see #1200), upgrade javassist to latest version which is JDK9 compatible to suppress this warning : 
`WARNING: Illegal reflective access by javassist.util.proxy.SecurityActions (file:.../play-1.x/framework/lib/javassist-3.21.0-GA.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)`


tomparle:
Thanks Andrei.
My bad not to have seen some test failed, I'm going to check and fix this as soon as possible !

tomparle:
OK, it took me more time than expected to understand the issue and finally fix it, but I'm glad to have found it out !
The class `PropertiesEnhancer` seems to have contained a workaround to fix an issue in javassist previous version :
 `// protected classes will be considered public by this call
 && Modifier.isPublic(ctField.getDeclaringClass().getModifiers());`

This call to `Modifier.isPublic()` should have returned false for a protected class (from javassist flags), but was returning true up to javassist version 3.21, hence the previous workaround in Play with the previous comment.
Starting from Javassist 3.23, the method `Modifier.isPublic()` correctly returns false for a protected class, and returns true for the method `Modifier.isProtected()`, hence the proposed fix.

PR #1252, opened 2018-08-03
Update Play source for build compability with JDK9+ 

tomparle:
Hi,

after the work for JDK9 compatibility of #1200, I wanted to check if Play could be built with Java 9 or 10.
This pull request contains the modifications I made to make this compatibility possible : it fixes some deprecated or missing methods, and upgrade a few dependencies, so it prepares Play for future builds with newer JDKs.

**Details :** 
- removed "source" and "target" from all ant build files so the user can set it through an environment varible
- libraries update : `gson 2.8.0 => 2.8.5, HikariCP 2.7.9 => 3.2.0, cglib 3.2.4 => 3.2.7`
- groovy 2.4.15 => 2.5.1. The previous groovy-all jar package is no more maintained, it is now just a POM which references a like a dozen of jars that Play does not need (they say this is the result of the work in progress for JDK9 full modularization). It seems that only groovy and groovy-xml are needed.
- hibernate-entitymanager has been removed since it is now included in hibernate-core. See https://github.com/hibernate/hibernate-orm/blob/5.2/migration-guide.adoc#hibernate-entitymanager-merged-into-hibernate-core
- in `GroovyTemplate.java`, the method named with an underscore generated a compilation error and has been removed. The hard part for me was to figure out where to replace these calls still existing in compiled Groovy templates.

Additional note : the test DataBinding failed on my machine because the language cookie seems to be ignored for some reason. I will see if it is ok on Travis.


tomparle:
Additional note : after a syntax test (using the "private" identifier on a method implementation in an interface, a new JDK9 feature), the compilation failed, so : 
- I also updated the libraries JDT and ASM
- I improved the java source detection mechanism to use the value specified in the `application.conf` file. Current compatible versions, that match JDT's, are now 1.8, 9 and 10, and I added an explicit error message if the requested source version is not available.

framework/build.xml
```diff
@@ -64,7 +64,7 @@
 
     <target name="compile" description="compile without cleaning">
         <mkdir dir="classes"/>
-        <javac encoding="utf-8" srcdir="src" destdir="classes" debug="true" source="1.8" target="1.8">
+        <javac encoding="utf-8" srcdir="src" destdir="classes" debug="true">
```
asolntsev: Why was it needed? I guess it's still a good idea to explicitly declare that Play itself is runnable with Java8 (even if somebody occasionally builds it with Play9+).

framework/conf/dependencies.yml
```diff
@@ -0,0 +1,121 @@
+
```
asolntsev: Why is this file needed? I don't see that it's used somewhere.

framework/build.xml
tomparle: @asolntsev thank you for your comments !
I wanted to give more freedom to choose the source and target versions by specifying them as environment variables, especially to test compatibility with building with jdk9 or 10.
Indeed it may be a good idea to keep target and sources at "1.8" for now, I'll make these changes.

framework/conf/dependencies.yml
asolntsev: @tomparle There is two files:
1. `framework/dependencies.yml` - it's actually used
2. `framework/conf/dependencies.yml` - it's not used. You added this file in your PR. 

Yes, file `framework/conf/dependencies.yml` is not used and should be deleted.

framework/src/play/templates/GroovyTemplate.java
```diff
@@ -521,7 +508,7 @@ public String __reverseWithCheck_absolute_false(String action) {
             return __reverseWithCheck(action, false);
         }
 
-        private String __reverseWithCheck(String action, boolean absolute) {
+        private static String __reverseWithCheck(String action, boolean absolute) {
```
asolntsev: Why you made it static? It's private method, it cannot be used from outside. Please make it non-static back.

tomparle: Thanks for the feedback and having taken the time to test for multiple versions !
I made this method static because my editor automatically notifies me when a method could be declared static (as an utility method), which I did here because it does not use any instance field or method. 
No problem if you still prefer to make it non-static back !

asolntsev: @tomparle Ok, I know that editor notification. I also used it many years ago.
Now I am pretty sure that `static` is a bad practice in most cases. For many reasons. The fact that the method doesn't use any fields is not the key factor.

tomparle: @asolntsev thanks for the feedback. I removed the static identifier from the method. I'd also be glad if you could share me some reasons of why it can be a bad practice, or give me a link about it !

Also, I just found out a regression that I did not see before, in the `#{form}` tag of the CRUD module (this module is separated from the framework, which is why I did not see it at first). This has been fixed.

I think it would also be neat to add the following kind of migration note about the removed method "_" in Groovy templates - even if I do not think it has been much used anywhere except from the framework itself (how do we add this to the future release notes ?) :
The shorthand method `_(String className)` to load a class in Groovy templates has been removed since the underscore is now a reserved keyword in JDK9 and later versions. Please use the method `__loadClass(String className)` instead.

asolntsev:
@tomparle Thank you! Now PR is good, going to merge it.

1. I don't know how to add it to Play 1 release notes. @xael-fry typically does it somehow.
2. it's a pity that CRUD module is not covered by automated tests :(
3. Here are a couple of links about static evilness for well-known guy Yegor Bugayenko. I am not always agree with his points, but these are good enough for initial reading:
- https://twitter.com/yegor256/status/753349557191180288
- https://www.yegor256.com/2015/02/20/utility-classes-vs-functional-programming.html
- https://www.yegor256.com/2017/02/07/private-method-is-new-class.html

tomparle:
Thanks for the tips @asolntsev  ! 
1. Ok, let's hope @xael-fry will be notified of the request to update the documentation. If not, I will keep in touch with him before the next release.
2. I agree. I will create one next time I deal with views
3. Thanks a lot for these links ! Interesting to read indeed.

framework/src/play/templates/GroovyTemplate.java
tazmaniax: @tomparle will this restriction of JDK9 on the Groovy templates have a knock on effect on https://github.com/codeborne/play-fastergt?

tomparle: @tazmaniax indeed, I proposed some changes to the `form` and `types` tag implementation, see https://github.com/codeborne/play-fastergt/pull/1
As I do not use this tag implementation, I wish someone who does could test it in his app.

tazmaniax: @tomparle I'll test shortly, thx! Out of interest is there a reason why you don't use this rendering implementation as it would appear to be an improvement over the standard implementation?

tomparle: @tazmaniax First, thank you for make me discover this implementation I was not aware of. I guess it's the main raison I was not using it ! But if indeed this is an improvement I will try it for sure.
Another issue is that I have already customized heavily CRUD tags to support additional features, and I guess I would have to merge it with faster GT implementation which would require some work. Anyway, please let me know of the test results !

tazmaniax: I've also done some customisation of the CRUD tags and it would be interesting to maybe coalesce them into this project if they would be of value to others

tomparle: Good idea ! I will see that after I test to integrate them on my project, and tell you if I can add some features useful to others.

xael-fry:
@tomparle  witch page of the documentation do you want to update.
You can directly update the doc cf https://github.com/playframework/play1/blob/master/documentation/manual/configuration.textile for java source.. should be done in the PR

tomparle:
@xael-fry I wanted to update the java compatible versions for the `java.version` property, adding `9` and `10` as compatible versions. Unfortunately I was unable to add it to the PR because recent merges in the master created conflicts I do not know how to resolve (conflict with dependencies upgrade with another PR). 
Additionally, I also wanted to precise the following migration notes :
`The shorthand method _(String className) to load a class in Groovy templates has been removed since the underscore is now a reserved keyword in JDK9 and later versions. Please use the method __loadClass(String className) instead`

xael-fry:
@tomparle ok so for ```java.version ``` I add it.
The 2nd point will be handle in #1261

tomparle:
Thanks @xael-fry !

tomparle:
Thank you for this very kind comment @jvosloo !
The :beers: will be a pleasure :wink:

Issue #1263, opened 2018-09-07
upgrade hibernate to 5.3.x


dreamingo88:
is there any plan to upgrade hibernate to 5.3.x

tomparle:
Hi,
I do not think there is such a plan currently, but libraries are regularly upgraded.
Is there any specific purpose for this upgrade ? What is your need ?
By the way, I think you should be able to use your own Hibernate lib by specifying it in your `dependencies.conf` file.

asolntsev:
@tomparle Unfortunately, it's not possible. The problem is that Play1 uses a "hacked" version of hibernate.jar

See files:
* https://github.com/playframework/play1/blob/master/framework/patches/hibernate-5.2.10-patch-play.patch
* https://github.com/playframework/play1/blob/master/framework/lib/hibernate-core-5.2.10.patched.jar

So, in order to update hibernate, you need to checkout sources, apply the patch, build a new hibernate jar and commit it to Play1. It's not impossible, but it's weird. :(

tomparle:
@asolntsev oh, I forgot about the Hibernate patch, my bad. Thank you for pointing this out and for the explainations to patch it manually.

flybyray:
We already use the newest hibernate with play.
We need to use hibernate feature to use SecondaryTables to integrate with legacy databases which split up data into multiple tables (without using onetoone workarrounds). the legacy databases should never be modified (and if possible hibernate should even respect that in ddl update mode).
actually there was a bug in hibernate [HHH-8805](https://hibernate.atlassian.net/browse/HHH-8805) for secondary tables which is fixed only in newest version.

asolntsev:
@flybyray Can you describe how exactly you did it? How is it possible to use latest hibernate.jar in a play application?

flybyray:
scm checkout hibernate sources, applying patches (some need adjustments for the current source of hibernate but not difficult, there are just line movements), than build hibernate. voila there is the jar. i am not sure if i have time to precisly describe it.

flybyray:
Ok I will provide the details. In that case i struggled with the current state of Play1 dependencies. It looks a little bit confused. https://github.com/playframework/play1/issues/1265
The hibernate-entitmanager is not referenced in dependencies yml.

flybyray:
That would be the current patch file https://gist.github.com/flybyray/35c1cde8655e65a49e4e14296c74c5a5
for hibernate 5.3.6

[...]

PR #1266, opened 2018-09-12
sample upgrade for issue #1263

flybyray:


flybyray:
I just checked the other offical dependencies of hibernate 5.3.6.Final
image

I think we need to adjust some other dependencies in play1 too.

xael-fry:
Merged thanks @flybyray

PR #1269, opened 2018-09-26
Add BigInteger binding

### LINKED ISSUE #1268: Fixture doesn't load data for BigInteger fields

1.5.1

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. Fixture for model with BigInteger field will load

### Actual Behavior

1. Values assigned to BigInteger fields are null

wytrych:
There seems to be most basic types in the Binder but
BigInteger was missing, even though BigDecimal was present.
Without the BigInteger binding fixtures for models using this type
will not load.

Fixes #1268

### DIFF
--- a/framework/src/play/data/binding/Binder.java
+++ b/framework/src/play/data/binding/Binder.java
@@ -14,6 +14,7 @@ import java.io.File;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.*;
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.text.ParseException;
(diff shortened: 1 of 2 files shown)


asolntsev:
@wytrych Thank you for the PR, but could you also add a test case to `BinderTest`?

wytrych:
@asolntsev test added. Let me know if it's ok and if so I will squash te commits.

wytrych:
Branch squashed.

asolntsev:
@wytrych thank you! Merged.

Issue #1262, opened 2018-08-30
Play1.4.3 and newer not allow to override default FastTag.class (empty namespace) using @FastTags.Namespace("")


anhtuan84:
### Play1.4.3 and newer not allow to override default FastTag.class (FastTag with empty namespace ) `@FastTags.Namespace("")`
I can not override defaut fastTag affter the fix : https://play.lighthouseapp.com/projects/57987/tickets/2040

In my application I override  default #{filed} tags to handle some new html attributes with the code like this:
```
@FastTags.Namespace
public class CustomTag extends FastTags {
    public static void _field(Map<?, ?> args, Closure body, PrintWriter out, ExecutableTemplate template, int fromLine) {
...
  }
}
```
From play-1.4.3 that does't works. Workaround solution is adding namespace to my tag `@FastTags.Namespace("custom")` but I have to change many template that use #{field}  to #{custom.filed}. I'm not sure that is a incident after fix issue: https://play.lighthouseapp.com/projects/57987/tickets/2040 or we don't want to allow override default FastTag

### Play Version (1.4.3 ) 
https://github.com/playframework/play1/compare/1.4.2...1.4.3
`GroovyTemplateCompiler.java:381`
Suggestion:
Change from
```
List<Class> fastClasses = new ArrayList<>();
fastClasses.add(FastTags.class);
 fastClasses.addAll(Play.classloader.getAssignableClasses(FastTags.class));
```
to 

```
List<Class> fastClasses = new ArrayList<>();
fastClasses.addAll(Play.classloader.getAssignableClasses(FastTags.class));
fastClasses.add(FastTags.class);
```

anhtuan84:
fixed https://github.com/thunderbird84/play1/commit/0b1ee22daa227b72a085c38a8e92fe2b8973ef47

xael-fry:
Merged in master

PR #1279, opened 2018-11-09
Allow to override default fast tags with empty namespace

tuantla:
# Allow to override default fast tags with empty namespace
https://github.com/playframework/play1/issues/1262


asolntsev:
@tuantla Is there a chance that you could add a test demonstrating the fixed functionality?

anhtuan84:
added test the case

PR #1292, opened 2019-01-18
Java 11 (JDT update)

sbeigel:
Get ready for Java 11: JDT updated, option for Java source version 11 added


sbeigel:
Ah, I'm afraid it's not that simple :) I think we have to update ASM to version 7.0 and to (re-)patch javaflow to make this work...

tomparle:
Thank you for the contribution Sebastian !
Indeed, the latest version of asm offers better support for the JDK 11, with some welcomed fixes, so it's a good thing to have upgraded it too !

sbeigel:
It's all green for me now, Java 11 source ist (pre-)compiled successfully, my web apps do work. I think together with my other PRs (mainly the hibernate 5.4 update) we should be ready for Java 11 :)

I'd appreciate if you could test a little bit as well and if it's all good, merge this stuff so we can all move to Java 11 ;)

sbeigel:
To avoid conflicts, I rebased on master after the hibernate 5.4 changes got merged.

xael-fry:
Merged and pushed in master
Thanks all @sbeigel @asolntsev ...

PR #1293, opened 2019-02-06
HikariCP: config property for init connection sql added

sbeigel:
Hikari supports setConnectionInitSql() to register custom SQL to initialize a new connection. This is handy to configure the character set for the connection (e.g. "SET NAMES utf8mb4" in MySQL to make emojis etc. work ;)
I added a play config option to setup this property.

### DIFF
--- a/documentation/manual/configuration.textile
+++ b/documentation/manual/configuration.textile
@@ -527,6 +527,12 @@ c3p0 is very asynchronous. Slow JDBC operations are generally performed by helpe
 Default: @3@
 
 
+h3(#db.pool.connectionInitSql). db.pool.connectionInitSql
+
+Sets the SQL string that will be executed on all new connections when they are created, before they are added to the pool (HikariCP only).
+
+Default: none
+
 h2(#evolutions). Database evolutions
 
 
--- a/framework/src/play/db/hikaricp/HikariDataSourceFactory.java
+++ b/framework/src/play/db/hikaricp/HikariDataSourceFactory.java
@@ -36,6 +36,10 @@ public class HikariDataSourceFactory implements DataSourceFactory {
     ds.setLoginTimeout(parseInt(dbConfig.getProperty("db.pool.loginTimeout", "0"))); // in seconds
     ds.setMaxLifetime(parseLong(dbConfig.getProperty("db.pool.maxConnectionAge", "0"))); // in ms
 
+    if (dbConfig.getProperty("db.pool.connectionInitSql") != null) {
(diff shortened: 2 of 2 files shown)


tazmaniax:
@sbeigel just looking at the [Connector/J 5.1 documentation](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-charsets.html) it explicitly warns...
> Do not issue the query SET NAMES with Connector/J, as the driver will not detect that the character set has been changed by the query, and will continue to use the character set configured when the connection was first set up.

and even recommends by default not to specify the connection string parameter `characterEncoding=UTF-8` and just rely on the server setting...
> The driver automatically uses the encoding specified by the server. For example, to use the 4-byte UTF-8 character set with Connector/J, configure the MySQL server with character_set_server=utf8mb4, and leave characterEncoding and connectionCollation out of the Connector/J connection string. Connector/J will then autodetect the UTF-8 setting.

but curious if you have a different experience

tazmaniax:
Related to this I've raised an issue https://github.com/playframework/play1/issues/1294 and pull request https://github.com/playframework/play1/pull/1295 to remove the current character encoding and collation defaults hardcoded by Play which are not needed and also sets the collation to "utf8_general_ci" which might not be desired especially if "utf8mb4" is configured on the server.

sbeigel:
> but curious if you have a different experience

Sometimes you can't (easily) change the server settings, so I was looking for another solution. And executing "SET NAMES" for every new connection seems to work, I can read and write extended unicode chars (emojis) without problems.

But this is only one (my current) usecase for this PR, so even if it's not the recommended or even a good solution for the utf8mb4 problem, having this setting in the play config is still useful!

tazmaniax:
> Sometimes you can't (easily) change the server settings, so I was looking for another solution. And executing "SET NAMES" for every new connection seems to work, I can read and write extended unicode chars (emojis) without problems.

Right I see your point particularly if you're using MySQL 5.1.46 (and earlier) without character_set_server=utf8mb4, then because `characterEncoding=UTF-8` maps to utf8 which is really utf8mb3 then quite possibly executing "SET NAMES utf8mb4" is the only option. However I would still be concerned about sending utf8mb4 encoded characters to a database that hadn't necessarily been configured to store utf8mb4 characters in the first place.

> But this is only one (my current) usecase for this PR, so even if it's not the recommended or even a good solution for the utf8mb4 problem, having this setting in the play config is still useful!

Definitely agree and I'm sure it will come in handy for other things, thx!

xael-fry:
Merged in master
thanks

PR #1295, opened 2019-02-11
Remove MySQL connection charset property defaults

### LINKED ISSUE #1294: Does not make sense to default MySQL character set and collation properties on connection URL
1.5.2

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)
MacOS 10.14.4

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

### Library Dependencies
Connector/J v5.1.47

### Behaviour
Play currently [adds the following default parameters to the connection URL](https://github.com/playframework/play1/blob/cd4fca38a56ecc25561302911a79667b471cf827/framework/src/play/db/DBPlugin.java#L189):
* useUnicode=yes
* characterEncoding=UTF-8
* connectionCollation=utf8_general_ci

The [Connector/J 5.1 documentation on charsets](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-charsets.html) does not recommend these properties are set instead preferring...
> The character encoding between client and server is automatically detected upon connection (provided that the Connector/J connection properties characterEncoding and connectionCollation are not set).

As well if the server collation is not set to "utf8_general_ci" and that collation is not expected then unexpected results might occur.

In addition the [documentation](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html) for `useUnicode` states that...
> Should only be used when the driver can't determine the character set mapping, or you are trying to 'force' the driver to use a character set that MySQL either doesn't natively support.

I suggest these defaults are removed to allow the client to freely auto-detect the character set and collation from the server and avoid setting potentially unexpected and undesirable defaults.

tazmaniax:
## Fixes
Fixes #1294 

## Purpose
Removes unneeded default MySQL character encoding and collation properties

## Background Context
Latest Connector/J documentation recommends no character encoding and collation properties to be set by default so that these values to be set by the server while not precluding these properties from being overriden by the client if necessary

### DIFF
--- a/framework/src/play/db/DBPlugin.java
+++ b/framework/src/play/db/DBPlugin.java
@@ -184,15 +184,9 @@ public class DBPlugin extends PlayPlugin {
                 String name = m.group("name");
                 String host = m.group("host");
                 String parameters = m.group("parameters");
-
-                Map<String, String> paramMap = new HashMap<>();
-                paramMap.put("useUnicode", "yes");
-                paramMap.put("characterEncoding", "UTF-8");
-                paramMap.put("connectionCollation", "utf8_general_ci");
-                addParameters(paramMap, parameters);
                 
                 dbConfig.put("db.driver", "com.mysql.jdbc.Driver");
-                dbConfig.put("db.url", "jdbc:mysql://" + (host == null ? "localhost" : host) + "/" + name + "?" + toQueryString(paramMap));
+                dbConfig.put("db.url", "jdbc:mysql://" + (host == null ? "localhost" : host) + "/" + name + "?" + parameters);
                 if (user != null) {
                     dbConfig.put("db.user", user);
                 }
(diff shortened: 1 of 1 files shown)


tazmaniax:
@xael-fry I had a look through the Travis log a couple of times and I don't think the issues had anything to do with the changes. When I compared the failed raw log with another successful log the issues seem to be more with resolving the dependency `play-codeborne->guice 1.9`. Could it just have been a bad run?
screenshot 2019-02-24 12 09 29

asolntsev:
@tazmaniax Yes, it seems it was just a bad run. 
@xael-fry Can you restart the build? I see a warning that build was migrated from travis.org to travis.com ...

tazmaniax:
For what its worth, I've been using this change in production since I submitted the pull request

tazmaniax:
@xael-fry can the build be restarted so we can see if the problem is repeated?

xael-fry:
@tazmaniax I cannot restart the build since travis move to the new url, can force the last commit to trigger the build again?

tazmaniax:
@xael-fry sorry, is that something you want me to do? Can you clarify? Should I just make a dummy commit to force the build?

xael-fry:
Yes a dummy commit or a amend and force push. Thanks

tazmaniax:
Horaah, the build worked :)

xael-fry:
@tazmaniax Thanks
Merged in master

PR #1308, opened 2019-06-10
Adds exit codes to play deps command

SomelifeDev:
## Purpose

This commit adds a non-zero exit code to the "play deps" command, allowing scripts to check if the dependencies were properly fetch or not.

## Background Context

We needed to be able to check how the dependencies fetching went for our integration process


xael-fry:
Merged in master
Thanks @asolntsev @SomelifeDev

Issue #1038, opened 2016-12-01
HTML escape in internationalized message


Rotkaf:
Hi,

Recently I've noticed some strange thing in template mechanism (XSS bug?).
I mean: when I write an action like this:

```
public static void someAction(String xssAtack)
{
	// xssAtack == "<script>alert(1)</script>";
        render("main.html", xssAtack);
}
```

and in messages file: `myText=XSS %s`
	
and in main.html: `&{'myText', xssAtack}`
	
this causes showing the alert! Shouldn't message parameter be auto HTML-escaped? I think it can be a **security hole in PF.** Am I right?

Of course, when I do in main.html: `${xssAtack}`
	
variable "xssAtack" will be escaped correct, according to documentation: 

> All dynamic expressions are escaped by the template engine to avoid XSS security issues in your application.

I know of course that I can escape every String I put in renderArgs by hand. If I have simple Strings is not a problem but what if I have a "big" object with objects as field and so on...?

So my question is: shouldn't the template engine do escaping the internationalized message as well as expressions?

asolntsev:
@Rotkaf I agree. Play framework should escape parameters of messages.

@xael-fry What do you think?
Probably we could make it at least configurable (not to break existing functionality)?

asolntsev:
@Rotkaf @xael-fry I have changed my mind :)
Actually it's not a security hole because all messages come from `messages.*` files that are located **inside** application code and never come from user. So, hacker **cannot** put `<script>` there.

sergey-moshnikov:
@asolntsev Why isn't it a security hole? Basically, it allows unescaped HTML to sneak into the page content via using parameterised messages. It's true that the message text comes from `messages.*` file, but the parameter values can come directly from the user.

tomparle:
@sergey-moshnikov I do not think too that it is a security hole, because no parameter value *should* come directly from the user. Instead, it would be the responsibility of the controller, for example, to escape the user input properly. Except this case, messages are indeed controlled by the application, so no arbitrary value should get in.

sergey-moshnikov:
Not to start a debate, but there's escaping in `${}` by default, why `&{}` should be less safe in that regard?

tomparle:
Thanks to your explanation about the wrong pull request, the multiple make sense !
Please let me quote your points from the PR in this discussion so everybody can read them :

> However, I tend to disagree with your reasoning that it's not a security issue, for 2 reasons:
[...]

PR #1318, opened 2019-08-21
HEALTH-198 prevent XSS through localized message parameters

sergey-moshnikov:
# Motivation

As stated in HEALTH-198, there's a XSS found in production, which basically executes as follows. 
Having a localized message tag with parameters in the html template, e.g. 
```
localized.message.key = This is a message parameterized with %s
param = "<script>alert(document.domain)</script>"
```
```
&{'localized.message.key', param}
```
would cause `param` to go unescaped into the result string and be injected into the HTML page. 

# Solution
The most general solution is to force message parameters be escaped in `GroovyTemplate.ExecutableTemplate.__getMessage()`.


tomparle:
Thanks for this pull request, but as stated before in https://github.com/playframework/play1/issues/1038, I am still not sure that fixing this is relevant, since :
- no user input should be directly used in messages, but properly escaped on the controller side
- the application may want to pass some html code as message parameters for easy styling

Moreover, this pull request contains quite a lot of modifications that are not directly related to the message parameter issue, but there are a lot more like additional logging and other modifications that may be valuable but that I think should be be moved into another dedicated request.
This is just my opinion, so let's hear about what other guys are thinking !

sergey-moshnikov:
Yeah sorry about that, that pull request was meant for Nosto/play1 fork, not the main repository. 
However, I tend to disagree with your reasoning that it's not a security issue, for 2 reasons: 
1) It's really easy to pass unescaped user input inadvertently, and I'd expect a framework to handle it for me; 
2) Play! framework does escape `${}` strings in templates, so it's quite natural that you'd expect similar behaviour from `&{}` syntax. 

Passing html code in parameters can be handled with calling `.raw()` explicitly. 
Anyways, as I said, I created this pull request by mistake.

PR #1319, opened 2019-08-22
hEALTH-198 Fix to the previous fix :)

sergey-moshnikov:
We can't apply __safeFaster() to all arguments; it should only affect `String`s


sergey-moshnikov:
Sorry, same mistake

PR #1329, opened 2020-02-05
fix cc mailaddress display name Garbled on multibyte character

ken-takagi:
# Pull Request Checklist

* [ ] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [ ] Have you updated the documentation?
* [ ] Have you added tests for any changed functionality?

# Helpful things

## Fixes

Fixes #xxxx

## Purpose

What does this PR do?

## Background Context

Why did you take this approach?

## References

Are there any relevant issues / PRs / mailing lists discussions?

### DIFF
--- a/framework/src/play/mvc/Mailer.java
+++ b/framework/src/play/mvc/Mailer.java
@@ -579,7 +579,12 @@ public class Mailer implements LocalVariablesSupport {
             List<String> ccsList = (List<String>) infos.get().get("ccs");
             if (ccsList != null) {
                 for (String cc : ccsList) {
-                    email.addCc(cc);
+                    try {
+                        InternetAddress iAddress = new InternetAddress(cc);
+                        email.addCc(iAddress.getAddress(), iAddress.getPersonal());
+                    } catch (Exception e) {
+                        email.addCc(cc);
+                    }
                 }
             }
 


to-lz1:
Hello. We use this play1 for our production, and want this PR to be merged.

https://lighthouseapp.com/dashboard seems to be unavailable,
so we can't figure out how to contact the reviewer of this project ...

We really appreciate if you notice and make any response. Thank you.

tomparle:
Hi @to-lz1 

Play developers definitely still see what's happening here, and this is the right place to discuss about it !
I do not see anything preventing from this PR to be merged, but releases are indeed not frequent.
Maybe @xael-fry or @asolntsev could plan a next release soon ?

Meanwhile, you could also use a custom build of Play which would include this PR and use it. Tell us if you need some tips to do so.

asolntsev:
That's right, we do see pull requests and comments.

In this pull request, I would like to see a unit-test showing that the change does work. I expect to see the email sample that did need such a change.

Then @xael-fry could make a release.

to-lz1:
@tomparle @asolntsev 

Thank you very much for your quick response!

I was relieved to hear that Play developers still check here, and I agree with @asolntsev. I think it would be better to add some unit-test and email samples to this PR. I'll talk to @ken-takagi about that (in fact, he is my colleague).

to-lz1:
@asolntsev Hi, I took a look at this issue.

Without this change, mail address handling will be slightly different from what we expect.

```
jshell> import org.apache.commons.mail.SimpleEmail;
jshell> import javax.mail.internet.InternetAddress;

jshell> Email sampleMail = new SimpleEmail();
jshell> String multibyteCharAddress = "Eva Nováková <test-1@example.com>"

# current code's approach
jshell> sampleMail.addCc(multibyteCharAddress);
jshell> sampleMail.getCcAddresses().get(0)
==> "Eva Nováková" <test-1@example.com>

# this PR's approach
jshell> InternetAddress iAddress = new InternetAddress(multibyteCharAddress);
jshell> sampleMail.addCc(iAddress.getAddress(), iAddress.getPersonal());
jshell> sampleMail.getCcAddresses().get(1)
==> =?UTF-8?Q?Eva_Nov=C3=A1kov=C3=A1?= <test-1@example.com>
```

As we can see, in the second example, non-ASCII characters are encoded based on [RFC2047](https://datatracker.ietf.org/doc/html/rfc2047).

Without this, our mailer can't know how to decode recipients' addresses, especially when they include some multibyte characters. We sometimes get garbled addresses such as `������ <foo@example.com>` displayed.

I considered adding tests about this. But I think this problem is rather related to the behavior of [Apache Commons' Email class](https://github.com/apache/commons-email/blob/master/src/main/java/org/apache/commons/mail/Email.java). 

Moreover, in [this(12 years old) commit](https://github.com/playframework/play1/commit/1e5458ea876f008d32df5f783fb28cd7eeff9d51), other (From, ReplyTo, and Bcc) addresses already experienced this change. Probably this change list has no bad effect, doesn't it?

asolntsev:
@to-lz1 Yes, I understand that this test seems more like a test for Apache Commons' Email class. But not only. It verifies that Play _does use_ the right library. 
Anyway, it's better to have such a test than not to have.

yuba:
@asolntsev  
Now that the UT you pointed out has been ready, could you consider merging this PR?

asolntsev (approved):

asolntsev:
@yuba I clicked "Accept" in code review. The last word is usually from @xael-fry - could you please merge this pull request and inform us about release plans?

yuba:
@xael-fry How about including this PR in release 1.8.0?

framework/src/play/mvc/Mailer.java
```diff
@@ -540,76 +534,84 @@ public static Future<Boolean> send(Object... args) {
...
-                    email.addReplyTo(replyTo);
-                }
+            return Mail.send(email);
+        } catch (EmailException ex) {
+            throw new MailException("Cannot send email", ex);
+        }
+    }
 
+    static void setAddresses(Email email) throws EmailException {
+        email.setCharset("utf-8");
```
xael-fry, line 558: why forced the charset to utf-8 here ? in the upper code you only set the charset if define ?
Do you mean that a default value is need ?

xael-fry (changes requested):

framework/src/play/mvc/Mailer.java
yuba: @xael-fry  
This encoding fix was originally written, in line 543.  
This line appears to have been moved to the setAddresses () method as it relates to setting addresses.

PR #1332, opened 2020-06-03
Compatibility with JDK 13 and 14

tomparle:
# Pull Request Checklist

* [ ] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [ ] Have you updated the documentation?
* [ ] Have you added tests for any changed functionality?

# Helpful things

## Fixes

Fixes https://github.com/playframework/play1/issues/1324

## Purpose

Make Play apps runnable with JDK version up to 14.
See corresponding issue for details and migration steps for faster-gt module users.

PR #1335, opened 2020-06-29
groovy update (to be Java 14 compatible)

sbeigel:
Groovy 2.5.3 is not working with Java 14, so I updated to 2.5.12.


tomparle:
Hi,

you may want to have a look at #1332, which already included a Groovy update and other modifications.

sbeigel:
Ah, great! Didn't notice that PR! Thanks! I close this PR...

PR #1345, opened 2021-01-09
Update to Eclipse JDT Core to 3.24 to enable JDK 15 source (and update other 3rd party libraries)

tazmaniax:
## Purpose

Primarily to update Eclipse JDT Core to the latest version v3.24 that supports JDK 15 source level. As well upgrades a number of other 3rd party libraries.

I tested this with both JDK 1.8 and JDK 15 and some tests had to be updated slightly but otherwise all tests pass with some unexplained errors being recorded that I was not able to resolve, see
CleanShot 2021-01-09 at 17 22 17
I'm not sure these are show stoppers.

## References
Here are the console output for the two runs using JDK 1.8 and JDK 15
[run8-.txt](https://github.com/tazmaniax/play1/files/5791576/run8-.txt)
[run15-.txt](https://github.com/tazmaniax/play1/files/5791577/run15-.txt)


tomparle:
Thanks for the contribution Chris.
Did you experience the same test errors when running the tests in your local environment ?

I think we would be more confident in future builds if we could clean any errors if possible.

tazmaniax:
Hey @tomparle , all of the test errors were from running in my local environment but it was almost impossible to identify from the log files exactly what was causing those two errors I mentioned above. I did update other tests to resolve other failures and errors but those two finally eluded me. Interesting to see if others get the same errors and if they're able to diagnose the cause. I've been running this version now for a couple of weeks in production and haven't seen any issues but I release that's not definitive proof that all is well.

xael-fry:
As we add support for java 17 in #1366, we can close this one

Issue #1353, opened 2021-04-14
python3 support or workaround


jamesbroadshaw:
### Are you looking for help?

I have an app I am deploying to heroku and with the latest heroku-20 stack, support for python2 has been removed. I took a quick stab at trying to update play using 2to3 but ran into a bunch of issues (not being a python developer). Perhaps porting to python3 is too much work but is it possible to bundle in the python2 runtime locally for linux as it's done for Windows and use that by default? Open to any ideas and happy to put up a PR that's mostly broken for suggestions.

### Play Version (1.5.x / etc)

1.6.0 (python2)

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

heroku-20 stack based on Ubuntu 20.04 LTS (without python2)

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

openjdk 16 2021-03-16

### Library Dependencies

heroku-20

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. heroku create
2. deploy using buildpack for play 1
3. application deploys

### Actual Behavior

python2 has been removed from heroku-20 stack and running the initial 'play' command gets a syntax error as it is being run with python3 instead.

tomparle:
Hi @jamesbroadshaw , nice issue you found out !
I'am also using Heroku and after reading https://devcenter.heroku.com/articles/heroku-20-stack it seems that the easiest would be to use the `python2` executable instead of `python`, which could be either by : 
- updating the script files in Play to choose the path to the python bin to use
- updating the Heroku Play build pack so that `python` calls link to `python2` with a symlink for example

While the second option may seem easier to implement, I'm not sure yet what would be best. 
What do other people think about his ?

jamesbroadshaw:
Thanks for taking a look @tomparle  (and all the other fixes you've contributed to keep play working with modern Java!)

python2 is only present during the build stage (slug compiler) of the app, the actual runtime when the slug is deployed to the dyno does not have it. I made a new app (replacing the top of the play python script of  _#!/usr/bin/env python_ with _python2_ to get it to build), and on the startup of the dyno:

`2021-04-15T01:04:13.320317+00:00 app[web.1]: [heroku-exec] Starting
2021-04-15T01:04:13.367977+00:00 app[web.1]: /usr/bin/env: 'python2': No such file or directory
2021-04-15T01:04:13.441905+00:00 heroku[web.1]: Process exited with status 127
[...]

PR #1355, opened 2021-04-30
Play scripts upgrade to Python 3

tomparle:
Upgrade all Play commands to python 3 (they were originally written in python 2).

## Purpose
This allow to keep Play 1 up-to-date with the current status of Python, whose version 2 is now longer maintained since January 1st, 2020.
This also allows to use Play in modern environments without Python 2 support, like Heroku current stack.

## Fixes
fixes https://github.com/playframework/play1/issues/1353

## Notes
I helped the migration with `futurize` but there was still quite some work to do manually.
I tested most commands and a successful build to Heroku-20 stack as well (which only supports python 3).

## Migration notes
This is a **breaking change**. This PR removes Python 2 support and requires now Python 3.

You may want to remove some print "hello" lines in some modules of your project that uses Play. The syntax causes an error in Python 3, and those lines are just sample code from the module template. They can be removed safely.
If you have custom-made modules with python scripts, you should migrate them manually to Python 3 (see some hints here : https://docs.python.org/3/howto/pyporting.html)


xael-fry:
The python exec has not been updated (/python), Do we really need to keep it ? I thinking about removing it ?
What do you think ?
@tomparle @asolntsev @tazmaniax ?

asolntsev:
> The python exec has not been updated (/python), Do we really need to keep it ? I thinking about removing it ? What do you think ? @tomparle @asolntsev @tazmaniax ?

I personally haven't used `/python` for a long time, but how will work users who used to `play foo` commands?

xael-fry:
the wrapper is only for windows user, as it is the python.exe. So for windows user they will need to have it on their computer same for linux. or am I missing something ?

asolntsev:
@xael-fry Ah Windows... Then we can drop it. Who cares about Windows users. :)

tazmaniax:
Ah ok, I was confused about what you were referring to. As long as I can still call Play commands on MacOS and Linux then I'm good 😄

xael-fry (approved):

tomparle:
Sorry for the late answer and thank you for the merge !
Indeed, the /python folder does not seem mandatory, but since I do not use Windows I'm not sure 100%.
Thanks for all the feedback.

aleksandy:
Removing the /python folder will make play! useless out of the box in a Windows environment if the python is not installed. In my opinion, it is very bad for newbies if the getting started tips will fail because of lack of python.

xael-fry:
@tomparle can you rebase this pr to master ?

tomparle:
@xael-fry I spent quite some time to rebase properly (and fix my previous rebase mistakes), upgrade some recent changes to python 3 syntax, and fix some python libs which I did not setup properly due to some confusion with python system libs.

Finally the merge and the tests seems ok, except for the windows environment test suite which does not terminate. I do not use Windows for developing and does not know what's causing these tests to be suspend so some help could be useful.
Anyway I do not think this should be blocking for this merge.

Alexandermjos:
Has anyone tried this on Windows? I get "ModuleNotFoundError: No module named 'win32pdh'" when using python 3.9.1 on Windows. 
"include win32pdh" works fine in python 2.7.8 included in Play 1.5.2

rakix:
$ pip install pypiwin32

works fine Play 1.7 with Windows

Alexandermjos:
Thank you 👍 
Is this considered common knowledge or something to consider putting into the documentation?

PR #1358, opened 2021-06-17
Update the README, use Markdown instead of Textile

cies:
The README was getting out of date. Links to Lighthouse (which is down), links to Play2 resources, and fairly little about that Play1 actually is and why it exists.


JohannesBeranek:
@cies would you like to update this PR?

xael-fry:
done in https://github.com/playframework/play1/pull/1474

PR #1369, opened 2022-01-05
GH-1328: Added SameSite support for session cookie.

### LINKED ISSUE #1328: Support SameSite cookie (Strict, Lax, None)
### Are you looking for help?

From Google Chrome 80, they are planning to released from february, 2020, they will force SameSite=None for all Cookies if we want to use for third-party context. It seems to be reversed the result as before (SameSite=Lax by default). Please see: https://www.chromium.org/updates/same-site
Therefore, we need to update our cookies with declarative setting SameSite from Playframework Controller.

Playframework 1.5.0

### Expected Behavior
1. Support SameSite enum inside the Http.Cookie
2. SameSite enum supports three values: Strict, Lax and None

Alexandermjos:
# Pull Request Checklist

* [ ] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you referenced any issues you're fixing using [commit message keywords](https://help.github.com/articles/closing-issues-using-keywords/)?
* [ ] Have you updated the documentation?
* [ ] Have you added tests for any changed functionality?

# Helpful things

## Fixes

Fixes #1328 

## Purpose

**What does this PR do?**
This PR add support for SameSite attribute on the session cookie (https://owasp.org/www-community/SameSite) 

## Background Context

**Why did you take this approach?**
At first, I created PlayCookie.java which extended Cookie.java, and DefaultPlayCookie.java extending DefaultCookie and implementing PlayCookie because Netty does not support SameSite attribute. 
I realized that for this to work, I had to implement PlayServerCookieEncoder to override encode methods in ServerCookieEncoder and for the encoder to work I had to copy CookieUtil from Netty in to the Play source code as it was package private. It worked, but it was a lot of code copy and duplication.

I refactored the code and ended up with this solution. Instead of implement a new version of ServerCookieEncoder, I call the original encoder (in Playhandler) from Netty and add SameSite at the end of the encoded string if the SameSite attribute is defined.

The SameSite attribute can be defined in "application.conf" with "application.session.sameSite"
Can ble not defined, blank, "None", "lax", "strict". 
Not sure if casing is browser independent. "Lax" and "lax" resulted in "Lax" in Chrome, but "none" was only registered with capital N ("None").

## References

**Are there any relevant issues / PRs / mailing lists discussions?**
Not that IO

### DIFF
--- a/framework/src/play/data/validation/ValidationPlugin.java
+++ b/framework/src/play/data/validation/ValidationPlugin.java
@@ -159,7 +159,7 @@ public class ValidationPlugin extends PlayPlugin {
         if (Validation.errors().isEmpty()) {
             // Only send "delete cookie" header when the cookie was present in the request
             if(Http.Request.current().cookies.containsKey(Scope.COOKIE_PREFIX + "_ERRORS") || !Scope.SESSION_SEND_ONLY_IF_CHANGED) {
-                Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", "", null, "/", 0, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY);
+                Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", "", null, "/", 0, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY, Scope.SESSION_SAMESITE);
             }
             return;
         }
@@ -179,7 +179,7 @@ public class ValidationPlugin extends PlayPlugin {
                 }
             }
             String errorsData = URLEncoder.encode(errors.toString(), "utf-8");
-            Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", errorsData, null, "/", null, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY);
+            Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", errorsData, null, "/", null, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY, Scope.SESSION_SAMESITE);
         } catch (Exception e) {
             throw new UnexpectedException("Errors serializationProblem", e);
         }
--- a/framework/src/play/mvc/Http.java
+++ b/framework/src/play/mvc/Http.java
@@ -163,6 +163,10 @@ public class Http {
          * See http://www.owasp.org/index.php/HttpOnly
          */
         public boolean httpOnly = false;
+        /**
+         * See https://owasp.org/www-community/SameSite
+         */
+        public String sameSite;
     }
 
     /**
@@ -707,8 +711,8 @@ public class Http {
          * @param value
          *            Cookie value
          */
-        public void setCookie(String name, String value) {
-            setCookie(name, value, null, "/", null, false);
+        public void setCookie(String name, String value, String sameSite) {
+            setCookie(name, value, null, "/", null, false, sameSite);
         }
 
         /**
@@ -730,7 +734,7 @@ public class Http {
          *            cookie path
          */
         public void removeCookie(String name, String path) {
-            setCookie(name, "", null, path, 0, false);
+            setCookie(name, "", null, path, 0, false, null);
         }
 
         /**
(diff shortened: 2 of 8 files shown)


miller11 (approved):

framework/src/play/data/validation/ValidationPlugin.java
```diff
@@ -159,7 +159,7 @@ static void save() {
         if (Validation.errors().isEmpty()) {
             // Only send "delete cookie" header when the cookie was present in the request
             if(Http.Request.current().cookies.containsKey(Scope.COOKIE_PREFIX + "_ERRORS") || !Scope.SESSION_SEND_ONLY_IF_CHANGED) {
-                Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", "", null, "/", 0, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY);
+                Http.Response.current().setCookie(Scope.COOKIE_PREFIX + "_ERRORS", "", null, "/", 0, Scope.COOKIE_SECURE, Scope.SESSION_HTTPONLY, Scope.SESSION_SAMESITE);
```
asolntsev, line 162: Isn't it strange that we use SameSite setting which was introduced **for session cookie** also for some other cookies?

framework/src/play/mvc/Scope.java
```diff
@@ -32,6 +32,7 @@
             .equals("true");
     public static boolean SESSION_SEND_ONLY_IF_CHANGED = Play.configuration
             .getProperty("application.session.sendOnlyIfChanged", "false").toLowerCase().equals("true");
+    public static final String SESSION_SAMESITE = Play.configuration.getProperty("application.session.sameSite");
```
asolntsev, line 35: How about renaming it to "application.session.cookie.sameSite"?

framework/src/play/mvc/Http.java
```diff
@@ -743,15 +747,15 @@ public void removeCookie(String name, String path) {
...
          */
-        public void setCookie(String name, String value, String duration) {
-            setCookie(name, value, null, "/", Time.parseDuration(duration), false);
+        public void setCookie(String name, String value, String duration, String sameSite) {
+            setCookie(name, value, null, "/", Time.parseDuration(duration), false, sameSite);
         }
 
-        public void setCookie(String name, String value, String domain, String path, Integer maxAge, boolean secure) {
-            setCookie(name, value, domain, path, maxAge, secure, false);
+        public void setCookie(String name, String value, String domain, String path, Integer maxAge, boolean secure, String sameSite) {
```
asolntsev, line 754: I suggest using _enum_ instead of String.

asolntsev (changes requested):

framework/src/play/data/validation/ValidationPlugin.java
Alexandermjos: Maybe, idk. What do you suggest? null / ""? 
It uses the same settings as session cookie for httpOnly and Secure, so I thought that it should also use same settings as session for samesite.

framework/src/play/mvc/Scope.java
Alexandermjos: > How about renaming it to "application.session.cookie.sameSite"?

Sure, I can do that.

framework/src/play/mvc/Http.java
Alexandermjos: > I suggest using _enum_ instead of String.

I can look into this. 
Input in application.conf is string. What is best practise for mapping to enum? Creating enums based on config value in uppercase? How to deal with incorrect values?

framework/src/play/data/validation/ValidationPlugin.java
Alexandermjos: Updated PR

framework/src/play/mvc/Scope.java
Alexandermjos: Updated PR

framework/src/play/mvc/Http.java
Alexandermjos: Updated PR

Alexandermjos:
Updated PR with requested changes @asolntsev

framework/src/play/mvc/Http.java
```diff
@@ -1018,4 +1022,20 @@ public WebSocketFrame(byte[] data) {
 
     public static class WebSocketClose extends WebSocketEvent {
     }
+
+    public enum SAMESITE {
```
asolntsev, line 1026: It's probably a matter of taste, but I would prefer `SameSite` name. Upper case doesn't make it readable.

asolntsev (approved):
Otherwise ok, but I would suggest to rename `SAMESITE` -> `SameSite`.

Alexandermjos:
I have updated PR with changes proposed by @asolntsev. Changed "SAMESITE" enum to "SameSite"

PR #1384, opened 2022-02-15
[#1383] 🆕 Ability to disable database evolutions on a datasource basis

### LINKED ISSUE #1383: Ability to disable database evolutions on a datasource basis
It would be really convenient to be able to disable database evolutions on a datasource basis.
The configuration currently support a global `evolutions.enabled=false` parameter.
Such a parameter could be present per datasource: `db.datasourcename.evolutions.enabled=false`

holajsh:
## Fixes

Fixes #1383 

## Purpose

Add support for a new configuration parameter to allow disabling database evolutions per database.

## References

#1383

### DIFF
--- a/framework/src/play/db/Evolutions.java
+++ b/framework/src/play/db/Evolutions.java
@@ -392,6 +392,10 @@ public class Evolutions extends PlayPlugin {
         return "false".equals(Play.configuration.getProperty("evolutions.enabled", "true"));
     }
 
+    private static boolean isDatabaseEvolutionDisabled(String dbName) {
+        return "false".equals(Play.configuration.getProperty("db." + dbName + ".evolutions.enabled", "true"));
+    }
+
     private static boolean isModuleEvolutionDisabled() {
         return "false".equals(Play.configuration.getProperty("modules.evolutions.enabled", "true"));
     }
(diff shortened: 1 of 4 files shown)


framework/src/play/db/Evolutions.java
```diff
@@ -392,6 +392,10 @@ private boolean isDisabled() {
         return "false".equals(Play.configuration.getProperty("evolutions.enabled", "true"));
     }
 
+    private static boolean isDatabaseEvolutionDisabled(String dbName) {
```
asolntsev, line 395: You are doing double-negotiation here: "false" -> "Disabled" -> "!Disabled". The code may be slightly simplified.

holajsh: Of course, in fact I just kept the existing "style" to stay coherent with other methods :D
Can be changed if really required ;)

asolntsev: I am not sure. Probably not such important. :)

asolntsev (approved):

xael-fry (changes requested):
@holajsh can you update the documentation https://github.com/playframework/play1/blob/master/documentation/manual/evolutions.textile ?
Thanks

holajsh:
> @holajsh can you update the documentation https://github.com/playframework/play1/blob/master/documentation/manual/evolutions.textile ? Thanks

Sure, I have pushed something but I'm not really sure how to check it will display properly...

xael-fry (approved):

PR #1387, opened 2022-02-17
make default jdk 11

aleksandy:
Since https://github.com/playframework/play1/commit/6dabc5308399226cc7cb11d64fc12ebbd5529e8c play! is compiled and tested on jdk 11 and 17 in CI. So in my opinion it's good cause to migrate to jdk 11 in the next release.

### DIFF
--- a/documentation/manual/configuration.textile
+++ b/documentation/manual/configuration.textile
@@ -745,9 +745,9 @@ h3(#java.source). java.source
 
 Java source level, which overrides the @java.version@ system property. For example:
 
-bc. java.source=17
+bc. java.source=11
 
-Values:  @1.7@ (No longer supported since 1.5.0), @1.8@ (No longer supported since 1.6.x), @9@ (No longer supported since 1.6.x), @10@ (No longer supported since 1.6.x), @11@, @12@, @13@, @14@, @15@, @17@.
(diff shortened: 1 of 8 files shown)


asolntsev (approved):

aleksandy:
It looks like after a481dd3 there is no option to use java 8. Because eclipse.jdt.core was updated to 3.28.0 and starting from 3.27.0 it [requires java 11](https://git.eclipse.org/c/jdt/eclipse.jdt.core.git/commit/?id=32abdfb9c5100e8c2ff7669768b7d7fcc83c7e23).

documentation/manual/configuration.textile
```diff
@@ -745,11 +745,11 @@ h3(#java.source). java.source
 
 Java source level, which overrides the @java.version@ system property. For example:
 
-bc. java.source=1.8
+bc. java.source=17
 
-Values:  @1.7@ (No longer supported since 1.5.0), @1.8@, @9@, @10@, @11@, @12@, @13@, @14@, @15@, @17@.
+Values:  @1.7@ (No longer supported since 1.5.0), @1.8@ (No longer supported since 1.6.x), @9@ (No longer supported since 1.6.x), @10@ (No longer supported since 1.6.x), @11@, @12@, @13@, @14@, @15@, @17@.
```
xael-fry, line 750: I think that it should be 'No longer supported since 1.7.0' has the 1.6.0 still support java 8 ?

aleksandy: Only if it will be next release. Because code already is uncompilable with java 8.

xael-fry (approved):

PR #1411, opened 2022-05-03
Do not overwrite system property net.spy.log.LoggerImpl -> 1.7.1

sp00m:
`MemcachedImpl` forces the usage of `net.spy.memcached.compat.log.Log4JLogger` while it shouldn't.

Fixes https://github.com/playframework/play1/issues/1414.

### DIFF
--- a/framework/src/play/cache/MemcachedImpl.java
+++ b/framework/src/play/cache/MemcachedImpl.java
@@ -92,7 +92,7 @@ public class MemcachedImpl implements CacheImpl {
     public void initClient() throws IOException {
         String loggerImpl = System.getProperty("net.spy.log.LoggerImpl");
         if (loggerImpl == null) {
-            System.setProperty("net.spy.log.LoggerImpl", "net.spy.memcached.compat.log.Log4JLogger");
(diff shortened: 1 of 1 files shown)


framework/src/play/cache/MemcachedImpl.java
```diff
@@ -90,7 +90,10 @@ protected byte[] serialize(Object object) {
     }
 
     public void initClient() throws IOException {
-        System.setProperty("net.spy.log.LoggerImpl", "net.spy.memcached.compat.log.Log4JLogger");
+        String loggerImpl = System.getProperty("net.spy.log.LoggerImpl");
+        if (loggerImpl == null) {
+            System.setProperty("net.spy.log.LoggerImpl", "net.spy.memcached.compat.log.Log4JLogger");
```
sp00m, line 95: Should we default to `net.spy.memcached.compat.log.SLF4JLogger` instead?

sp00m:
If this could go to 1.7.1, would be 👌

holajsh:
Must definitely go into 1.7.1 as Memcached implementation is broken otherwise ;)

xael-fry (approved):

PR #1432, opened 2022-11-22
Better compatibility with Hibernate Envers

EddyVerbruggen:
## Purpose

When [Hibernate Envers](https://hibernate.org/orm/envers/) is triggered via an `@Audited` annotation, it attempts to insert a record in a related `AUD` table.

However, Play! assumes all related entities extend `JPABase`, which the `AUD` tables don't. This leads to a `ClassCastException` because `HashMap` is cast to `JPABase`.

By adding a check for `JPABase` here, we can avoid this problem. This effectively also checks for 'not null' which the previous implementation did.

### DIFF
--- a/framework/src/play/db/jpa/HibernateInterceptor.java
+++ b/framework/src/play/db/jpa/HibernateInterceptor.java
@@ -13,7 +13,7 @@ public class HibernateInterceptor extends EmptyInterceptor {
   public HibernateInterceptor() {
 
   }
-  
+
   @Override
   public int[] findDirty(Object o, Serializable id, Object[] arg2, Object[] arg3, String[] arg4, Type[] arg5) {
     if (o instanceof JPABase && !((JPABase) o).willBeSaved) {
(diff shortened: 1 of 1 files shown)


aleksandy:
@EddyVerbruggen, could you please provide a small test case demonstrating the problem? 
I've been using envers since play! 1.2.x and have never experienced this.

EddyVerbruggen:
Hi @aleksandy! I'd love to, but if you can use it without problems and I can't then it's probably something in the configuration of my entities (or something deeper down the stack) which probably makes it pretty time consuming to come up with a good test case.

So before I do, can I ask you to consider the change on merit (outside the context of Envers)?

I mean, checking for `instance of X` before trying to cast to `X` (instead of the current 'not null' check) is arguably better anyway. And it fixes the Envers compatibility along the way for my app. 🙏

aleksandy:
@EddyVerbruggen, I agree your point. But unfortunately I don't have enough rights for it.

@asolntsev, @xael-fry, could you approve and merge PR?

EddyVerbruggen:
Hi friends, at [Combidesk](https://combidesk.com) we're huge fans of Play!(!). We're using Play 1 for a few of our backend systems and would love to have this PR merged and released.

Is there anything we can do to help drive this forward? As OSS maintainers ourselves we know it may be hard to find time and prioritise unpaid work, so we'd be happy to do a decent donation for your efforts as well.

Cheers, and keep up the great work!

asolntsev (approved):
Lgtm

xael-fry (approved):

PR #1439, opened 2023-01-09
Fix class newinstance deprecation

tazmaniax:
Fixes deprecation with Class.newInstance()

Also fixes a deprecation with Runtime.exec().


xael-fry:
will be handle in https://github.com/playframework/play1/pull/1511 (rebase  + squash + add one missing)

PR #1448, opened 2023-07-14
Updated 3rd party dependencies

tazmaniax:
All tests pass


tazmaniax:
This is compiling successfully using JDK 20 so maybe the minimum JDK has been increased and the following libs must be reverted to 3.33.0. More investigation is needed.

    - org.eclipse.jdt -> org.eclipse.jdt.core 3.34.0
    - org.eclipse.jdt -> ecj 3.34.0

xael-fry (approved):

PR #1481, opened 2024-03-15
support for (custom) @Global unbinders added

sbeigel:
With the same approach and mindset as support for custom binders exists in Binder.java for ages, I patched Unbinder.java to support custom unbinders annotated with "@Global" and implementing "TypeUnbinder"


xael-fry:
rebase and push in https://github.com/playframework/play1/pull/1562

PR #1483, opened 2024-03-27
Support for JDK 21

osmanirosado:
## Purpose

Compile code written for JDK 21 to improve code readability using [Pattern Matching for switch Expressions and Statements](https://docs.oracle.com/en/java/javase/21/language/pattern-matching-switch-expressions-and-statements.html#GUID-E69EEA63-E204-41B4-AA7F-D58B26A3B232)


xael-fry:
done with #1538

PR #1501, opened 2025-06-24
Use setHeader instead of addHeader to prevent duplicate headers

churlevski-filip:
Use setHeader instead of addHeader to prevent duplicate headers


churlevski-filip:
not needed here

PR #1550, opened 2026-01-21
Cleanup

aleksandy:
* use toAray(IntFunction) instead of toArray(T[])
* optimize JavaExtensions#remove()
* simplify JavaExtensions#pad()
* simplify JavaExtensions#join()
* use primitives instead of wrappers
* use pattern matching instead of explicit cast
* use String.valueOf() and toString() for the objects stringification instead of concatenation with empty string


xael-fry (changes requested):
The MR is so called "cleanup" but it contains refactors,, fixes and all all parts off the code and not on a given part.
Please split your work in different MR, you can also use the conventional commit to have better description of the scope of the modification apply to.

Thanks

aleksandy:
@xael-fry, do you still have any comments on this PR? Could you merge it?

